PR #4, opened 2022-05-25
Update issue templates

gkellogg:
Standard and Use case issue templates.


pchampin (approved):

pchampin:
> @pchampin Templates more or less copied from RDF-star repo. Any other suggestions?

None that I can think of at the moment. Anyway, we can add other templates afterwards, as the need emerges.

PR #27, opened 2022-06-23
Adds a preliminary IANA section with placeholders.

### LINKED ISSUE #14: YAML LD media type and profiles
We need to apply for `application/ld+yaml` media type.
@ioggstream "I'm working on the YAML media type registration https://ietf-wg-httpapi.github.io/mediatypes/draft-ietf-httpapi-yaml-mediatypes.html and we already have an issue related to the registration of https://github.com/ietf-wg-httpapi/mediatypes/issues/8".

In addition, it seems clear to me that we'll need various YAML-LD **profiles**, eg as per #6 by @pchampin . See https://www.w3.org/TR/dx-prof-conneg/

### LINKED ISSUE #25: File extension
JSON-LD uses `.jsonld` as the (recommended) file extension as can be seen from the [IANA considerations and elsewhere in the specification](https://www.w3.org/TR/json-ld/#iana-considerations). So I would like to bring up an obvious question (because I am selecting the "main" file extension for a [project](https://github.com/twinbase/twinbase)):

## What will be the (recommended) file extension for YAML-LD?

I suggest `.yamlld` for consistency and the "no surprises" principle.

Previous materials on the topic:
- @gkellogg used `.ymld` in an example in this comment: https://github.com/json-ld/yaml-ld/issues/8#issuecomment-1140292068 
- https://github.com/json-ld/yaml-ld/issues/14 has used just plain `.yaml`.
  - (The upcoming IANA media type application also apparently has to define a file extension?)

gkellogg:
Fixes #14.
Fixes #25.

cc/ @anatoly-scherbakov @juusoautiosalo

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/27.html" title="Last updated on Jun 29, 2022, 9:22 PM UTC (d5c7055)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/27/58a20c3...d5c7055.html" title="Last updated on Jun 29, 2022, 9:22 PM UTC (d5c7055)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -10,11 +10,29 @@
     copyrightStart:    "2020",
     shortName:  "yaml-ld",
     edDraftURI: "https://json-ld.github.io/yaml-ld/",
-    github: "https://github.com/json-ld/yaml-ld/",
+    github: {
+      repoURL: "https://github.com/json-ld/yaml-ld/",
+        branch: "main"
+    },
     doJsonLd:   true,
     editors: [{
         name: "JSON-LD Community"
     }],
+
+    localBiblio: {
+      "I-D.ietf-httpapi-yaml-mediatypes": {
+        title: "YAML Media Type",
+        href: "https://datatracker.ietf.org/doc/draft-ietf-httpapi-yaml-mediatypes/",
+        publisher: "IETF",
+        date: "2022-06-23",
+        status: "WG Document",
+        authors: [
+          "Roberto Polli",
+          "Erik Wilde",
+          "Eemeli Aro"
+        ]
+      },
+    },
 
     // Cross-reference definitions
     xref: ["json-ld11", "json-ld11-api", "json-ld11-framing"],
@@ -229,7 +247,8 @@
         this document identifies constraints on YAML documents
         such that they can be used to represent JSON-LD documents.
       </p>
-        
+    </section>
+
   <section id="conformance">
     <p>A <a>YAML-LD document</a> complies with this specification if ...</p>
     <p class="ednote">Define <dfn>YAML-LD document</dfn> somewhere.</p>
@@ -270,7 +289,8 @@
   <section id="sec" class="informative">
     <h2>Security Considerations</h2>
 
-    <p>See the YAML media type registration.</p>
+    <p>See <a data-cite="JSON-LD11##iana-security">Security considerations in JSON-LD 1.1</a>.
+      Also, see the YAML media type registration.</p>
   </section>
 
   <section id="int" class="informative">
@@ -278,18 +298,19 @@
 
     <p>
       For general interoperability considerations on the serialization of
-      JSON documents in YAML, see [[YAML]].
+      JSON documents in YAML, see [[YAML]]
+      and the Interoperability consideration of application/yaml [[I-D.ietf-httpapi-yaml-mediatypes]]..
     </p>
         
   </section>
 
-  <section class="appendix normative">
+  <section id="iana" class="appendix normative">
     <h2>IANA Considerations</h2>
 
     <p>This section has been submitted to the Internet Engineering Steering
       Group (IESG) for review, approval, and registration with IANA.</p>
 
-    <h3>application/ld+json</h3>
+    <h3>application/ld+yaml</h3>
     <dl>
       <dt>Type name:</dt>
       <dd>application</dd>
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -248,5 +249,155 @@ <h2>Basic Concepts</h2>
...
+        directed graphs. Implementations of YAML-LD have been created for
+        <span class="note">FIXME</span>.
+      </dd>
+      <dt>Additional information:</dt>
+      <dd>
+        <dl>
+          <dt>Magic number(s):</dt>
+          <dd>Not Applicable</dd>
+          <dt>File extension(s):</dt>
+          <dd>.yamlld</dd>
```
ioggstream, line 372: Q: for json-ld the unique extension is .jsonld, right?

spec/index.html
```diff
@@ -248,5 +249,155 @@ <h2>Basic Concepts</h2>
...
+              <dt><code>http://www.w3.org/ns/json-ld#extended</code></dt>
+              <dd>To request or specify <a data-lt="extended document form">extended YAML-LD document form</a>.
+                <div class="ednote">
+                  This is a placeholder for specifying something like an
+                  <dfn data-lt="extended-document-form">extended YAML-LD document form</dfn>
+                  making use of YAML-specific features.
+                </div></dd>
+            </dl>
+            <p>All other URIs starting with <code>http://www.w3.org/ns/json-ld</code>
+              are reserved for future use by JSON-LD specifications.</p>
```
ioggstream, line 308: Do we need a  http://www.w3.org/ns/yaml-ld ?
Does URI reservation belong to media type registration?
If the answers are no, then remove.

```suggestion
```

spec/index.html
```diff
@@ -248,5 +249,155 @@ <h2>Basic Concepts</h2>
...
+            <!--p>Other specifications MAY create further structured subtypes
+              by using `+ld+json` as a suffix for a new base subtype, as in
+              `application/example+ld+json`.
+              Unless defined otherwise, such subtypes use the same
+              fragment identifier behavior as `application/ld+json`.</p-->
```
ioggstream, line 313: If we think this is the case, we need to file an issue and remove the commented text from the spec.

Otherwise, remove.

```suggestion
```

spec/index.html
```diff
@@ -248,5 +249,155 @@ <h2>Basic Concepts</h2>
...
+            <p>Other specifications may publish additional `profile` parameter
+              URIs with their own defined semantics.
+              This includes the ability to associate a file extension with a `profile` parameter.</p>
+            <p>
+              When used as a <a data-cite="RFC4288#section-4.3">media type parameter</a> [[RFC4288]]
+              in an <a data-cite="rfc7231#rfc.section.5.3.2">HTTP Accept header</a> [[RFC7231]],
```
ioggstream, line 319:
```suggestion
              in an <a data-cite="rfc9110#rfc.section.12.5.1">HTTP Accept header field</a> [[RFC9110]],
```

Probably, we could simplify this section, eg.

> When using this parameter in content negotiation (see Section 12.5 of RFC9110), its value MUST be
> enclosed in double quotes...

Q: is the procedure above specific to the profile parameter or it is generally valid for all media type parameters? In case, I can double check.

spec/index.html
```diff
@@ -248,5 +249,155 @@ <h2>Basic Concepts</h2>
...
+      <dt>Type name:</dt>
+      <dd>application</dd>
+      <dt>Subtype name:</dt>
+      <dd>ld+yaml</dd>
+      <dt>Required parameters:</dt>
+      <dd>N/A</dd>
+      <dt>Optional parameters:</dt>
+      <dd>
+        <dl>
+          <dt><code>profile</code></dt>
```
ioggstream, line 270: Q: this section is very long. To ensure it is aligned between yaml+ld and json+ld we could probably just reference
the parameter from either the ld+json registration or the json-ld specification document which was used for the registration of ld+json. WDYT?

ioggstream (changes requested):
Preliminary review

spec/index.html
```diff
@@ -248,5 +249,155 @@ <h2>Basic Concepts</h2>
...
+            <p>When processing the "profile" media type parameter, it is important to
+              note that its value contains one or more URIs and not IRIs. In some cases
+              it might therefore be necessary to convert between IRIs and URIs as specified in
+              <a data-cite="RFC3986#section-5.1">section 3 Relationship between IRIs and URIs</a>
+              of [[RFC3987]].</p>
+          </dd>
+        </dl>
+      </dd>
+      <dt>Encoding considerations:</dt>
+      <dd>See <a data-cite="RFC8259#section-11">RFC&nbsp;8259, section 11</a>.</dd>
```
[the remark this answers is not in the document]
gkellogg: How about https://datatracker.ietf.org/doc/draft-ietf-httpapi-yaml-mediatypes/ with an editor's note to update this when published? Or do you see YAML-LD being published before YAML has a registered media type? I think we should reference something.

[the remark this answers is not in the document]
gkellogg: Based on your below comment, we can add `I-D.ietf-httpapi-yaml-mediatypes` as a bib reference and use "See the Encoding considerations

Based on your below comment, we might just use "binary. (See [[I-D.ietf-httpapi-yaml-mediatypes]]).

spec/index.html
gkellogg: One might use `.yamld`, but consider this a placeholder subject to group concensus.

spec/index.html
gkellogg: We can certainly remove the paragraph, but the overriding specification is json-ld and yaml-ld uses this namespace. I don't think there's a hard-and-fast rule against creating additional namespaces, but as the spec is so aligned with JSON-LD I think it's probably best to keep such IRIs localized, at least for now.

spec/index.html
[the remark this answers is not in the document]
gkellogg: Please check out updates.

spec/index.html
gkellogg: I believe it's generally true about parameter values.

spec/index.html
gkellogg: Shortened, now.

spec/index.html
ioggstream: yamlld LGTM

ioggstream (approved):
LGTM

spec/index.html
ioggstream: @gkellogg OK to use http://www.w3.org/ns/json-ld for yaml-ld if needed.

spec/index.html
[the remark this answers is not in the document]
ioggstream: Great, thanks!

anatoly-scherbakov (approved):

spec/index.html
```diff
@@ -269,25 +289,133 @@ <h2>Basic Concepts</h2>
...
+              honors the profiles in the list which it recognizes,
+              and MUST ignore the profiles in the list which it does not recognize.
+              It is RECOMMENDED that profile URIs are dereferenceable and provide
+              useful documentation at that URI. For more information and background
+              please refer to [[RFC6906]].</p>
+            <p>This specification allows the use of the `profile` parameters listed in
+              <a data-cite="JSON-LD11##iana-considerations"> and additionally defines the following:</a>
+            </p>
+            <dl>
+              <dt><code>http://www.w3.org/ns/json-ld#extended</code></dt>
```
VladimirAlexiev, line 343: Should be http://www.w3.org/ns/yaml-ld#extended

PR #34, opened 2022-06-30
Preliminary rules for discussion

### LINKED ISSUE #8: Round-trip safe json-ld -> yaml-ld -> json-ld
As an <user with json-ld files>       … WHO
I want to <convert them to yaml-ld>  … WHAT
So that <they are round-trip safe>   … WHY

### LINKED ISSUE #15: char encoding: UTF-8 only?
https://ietf-wg-httpapi.github.io/mediatypes/draft-ietf-httpapi-yaml-mediatypes.html#name-yaml-and-json (@ioggstream)
"the following ones might have interoperability issues with JSON: non UTF-8 encoding, since YAML supports UTF-16 and UTF-32 in addition to UTF-8".

Can we demand that YAML-LD be in UTF-8 only?

Turtle allows only UTF-8 (https://www.w3.org/TR/turtle/#sec-mediaReg), so I see no loss for those other encodings

### LINKED ISSUE #46: Multiple documents in YAML
Should YAML-LD allow or prohibit multiple documents in YAML?
- Which YAML parsers support multiple documents?
- What are useful examples of using multiple documents?
- If we decide to use them in YAML-LD, how should they be represented? As RDF graphs?
- Below I formulate a positive use case, but I'm not quite certain we want this because of its complexity

PLEASE VOTE with :+1: or :-1: , thanks!

---

**Eg1**: multiple identical keys are forbidden by YAML linters.
But they are ok if they are in different documents.
Example by @ioggstream from https://github.com/json-ld/yaml-ld/issues/42#issuecomment-1173646556:
```
---
a: 1
...
---
a: 2
...
```
---
**Eg2**: YAML metadata followed by a markdown textual body is widely used in some blog/content management systems:
```
created: 2022-07-03
published: 2022-07-04
title: Frobnification
author: A. U. Thor
---
Frobnification was invented in prehistoric times.
It's a useful meta-process wherein...
```
---
As an information architect.
I want to be able to use multiple documents in YAML-LD.
So that I can transmit several closely related documents (graphs) together.

ioggstream:
## This PR

- fixes #8
- fixes #46

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/34.html" title="Last updated on Jul 6, 2022, 4:16 PM UTC (b73e250)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/34/13ff5f8...b73e250.html" title="Last updated on Jul 6, 2022, 4:16 PM UTC (b73e250)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -295,7 +295,7 @@
 
       <ul>
         <li>YAML supports different encodings, including UTF-8, UTF-16, and UTF-32.</li>
-        <li>YAML supports more data types than JSON.</li>
+        <li>YAML supports more native data types than JSON.</li>
         <li>the structure of a YAML document &mdash; that is, a named YAML representation graph &mdash;
         is a rooted, directed graph that can have cycles.</li>
         <li>YAML has the concept of stream, which is a sequence of documents.
--- a/spec/index.html
+++ b/spec/index.html
@@ -243,7 +243,7 @@
       <p>
         Since YAML is more expressive than JSON,
         both in the available data types and in the document structure
-        (see I-D.ietf-yaml-mediatypes),
+        (see [[I-D.ietf-httpapi-yaml-mediatypes]]),
         this document identifies constraints on YAML documents
         such that they can be used to represent JSON-LD documents.
       </p>
@@ -283,9 +283,156 @@
   <section id="basic-concepts" class="informative">
     <h2>Basic Concepts</h2>
 
-    <p>FIXME.</p>
-  </section>
+    <p>
+      To ease writing and collaborating on JSON-LD documents, it is a common practice
+      to serialize them as YAML.
+      This requires a registered media type, not only to enable content negotiation
+      of linked data documents in YAML, but even to define the expected behavior of
+      applications that processes these documents, including fragment identifiers and
+      interoperability considerations.
+    </p>
 
+    <p>
+      This is because YAML is more flexible than JSON:
+    </p>
+
+      <ul>
+        <li>YAML supports different encodings, including UTF-8, UTF-16, and UTF-32.</li>
+        <li>YAML supports more data types than JSON.</li>
+        <li>the structure of a YAML document &mdash; that is, a named YAML representation graph &mdash;
+        is a rooted, directed graph that can have cycles.</li>
+
+    <p>
+      The first goal of this specification is to allow a JSON-LD document to be
+      processed and serialized into YAML, and then back into JSON-LD, without
+      losing any information.
+
+      This is always possible, because a YAML representation graph can always represent
+      a tree, because JSON data types are a subset of YAML's, and because
+      JSON encoding is UTF-8.
+    </p>
+
+    <p data-format="markdown>Example: the JSON-LD document below
+
+    ```
+    {
+      "@context": "http://example.org/context.jsonld",
+      "@graph": [
+        {"@id": "http://example.org/1", "title": "Example 1"},
+        {"@id": "http://example.org/2", "title": "Example 2"},
+        {"@id": "http://example.org/3", "title": "Example 3"}
+      ]
+    }
+    ```
+
+    can be serialized as YAML as follows.
+    Note that entries 
+    starting with  `@` need to be enclosed in quotes or escaped because
+    `@` is a reserved character in YAML.
+
+    ```yaml
+    %YAML 1.2
+    ---
+    "@context": http://example.org/context.jsonld
+    \@graph:
+      -
+        "@id": http://example.org/1
+        title: Example 1
+      -
+        \@id: http://example.org/2
+        title: Example 2
+      -
+        '@id': http://example.org/3
+        title: Example 3
+    ```
+
+
+    </p>
+  </section>
+  <section id="specifications" class="normative">
+    <h2>Core Requirements</h2>
+
+    <p>
+      A YAML-LD document is a [[YAML]] document that can be interpreted as Linked Data [[LINKED-DATA]].
+    </p>
+    <p>
+      It MUST be encoded in UTF-8, to ensure interoperability with [[JSON]]. 
+    </p>
+    <p>
+      Comments in YAML-LD documents
+      are treated as white space
+      This behavior is consistent with other
+      Linked Data serializations like [[TURTLE]].
+      See Interoperability considerations of [[I-D.ietf-httpapi-yaml-mediatypes]]
+      for more details.
+    </p>
+    <p>
+      Since named anchors are a serialization detail,
+      such names
+      MUST NOT be used to convey relevant information,
+      MAY be altered when processing the document,
+      and MAY be dropped when interpreting the document as JSON-LD.
+    </p>
+    <p> 
+      A YAML-LD document MAY contain named anchors and alias nodes, 
+      but its representation graph MUST NOT contain cycles.
+      When interpreting the document as JSON-LD,
+      alias nodes MUST be resolved by value to their target nodes.
+    </p>
+    <p data-format="markdown">
+      Example: The following YAML-LD document
+      contains alias nodes for the `{"@id": "country:ITA"}` object:
+
+      ```yaml
+      %YAML 1.2
+      ---
+      "@context":
+        "@vocab": "http://schema.org/"
+        "countries": "http://publication.europa.eu/resource/authority/country/"
+      "@graph":
+      - &ITA
+        "@id": countries:ITA
+      - "@id": http://people.example/Homer
+        name: Homer Simpson
+        nationality: *ITA
+      - "@id": http://people.example/Lisa
+        name: Lisa Simpson
+        nationality: *ITA
+      ```
+
+      While the representation graph (and eventually the in-memory representation
+      of the data structure, e.g., a Python dictionary or a Java hashmap) will still
+      contain references between nodes, the JSON-LD serialization will not.
+
+      ```json
+      {
+        "@context": {
+          "@vocab": "http://schema.org/",
+          "countries": "http://publication.europa.eu/resource/authority/country/"
+        },
+        "@graph": [
+          {
+            "@id": "countries:ITA"
+          },
+          {
+            "@id": "http://people.example/Homer",
+            "full_name": "Homer Simpson",
+            "country": {
+              "@id": "countries:ITA"
+            }
+          },
+          {
+            "@id": "http://people.example/Lisa",
+            "full_name": "Lisa Simpson",
+            "country": {
+              "@id": "countries:ITA"
+            }
+          }
+        ]
+      }
+      ```
+    </p>
+  </section>
   <section id="sec" class="informative">
     <h2>Security Considerations</h2>
 
@@ -421,6 +568,97 @@
       <h3>Examples</h3>
       <p class="ednote">FIXME</p>
     </section>
+  </section>
+
+  <section id="faq" class="informative" data-format="markdown">
+    <p class="ednote">REMOVE THIS SECTION BEFORE PUBLICATION.</p>
+
+    <h3>FAQ</h3>
+
+    #### Why does YAML-LD not preserve comments?
+    <p class="ednote">
+    [[JSON]] (and hence [[JSON-LD]]) does not support comments,
+    and other Linked Data serialization formats
+    that support comments (such as [[TURTLE]])
+    do not provide a means to preserve them
+    when processing and serializing the document
+    in other formats.
+    The proposed behavior is thus consistent with
+    other implementations.
+
+    While YAML-LD could define a specific predicate for comments,
+    that is insufficient because, for example,
+    the order of keywords is not preserved in JSON, so the
+    comments could be displaced.
+    This specification does not provide a means for preserving
+    YAML comments after a JSON serialization.
+
+    ```yaml
+    # First comment
+    "@context": "http://schema.org"
+
+    # Second comment
+    givenName: John
+    ```
+    
+    transforming the above entry into a JSON-LD document
+    results in:
+    
+    ```json
+    {
+      "@context": "http://schema.org",
+      "givenName": "John"
+    }
+    ```
+
+    
+    #### Why does YAML-LD not extend the JSON-LD data model ?
+    <p class="ednote">
+    [[JSON]] only represents simple trees while [[YAML]] can support
+    rooted, directed graphs with references and cycles.
+
+    The above structures cannot be preserved when serializing
+    to JSON-LD and - with respect to cycles - the serialization
+    will fail.
+
+    Programming languages such as Java and Python already support
+    YAML representation graphs, but these implementations may behave
+    differently.
+    In the following example, `&value` references the value
+    of the keyword `value`.
+
+    ```yaml
+    value: &value 100
+    valve1:
+      temperature: &temp100C
+        value: *value
+        unit: degC
+    valve2:
+      temperature: *temp100C
+    ```
+
+    Processing this entry in Python, I get the following
+    structure that preserve the references to
+    mutable objects (e.g., the `temperature` dict)
+    but not to scalars (e.g., the `value` keyword).
+
+    ```python
+    temperature = { "value": 100, "unit": "degC" }
+    document = {
+      "value": 100,
+      "valve1": { "temperature": temperature },
+      "valve2": { "temperature": temperature }
+    }
+    ```
+
+    Since all these implementations pre-date this 
+    specification, some more interoperable choices include the following:
+    
+    * to forbid cycles in YAML-LD documents
+    * to consider all references in YAML-LD as static, 
+      i.e., a shorthand way to repeat specific patterns
+
+    </p>
   </section>
 
  </body>
--- a/spec/index.html
+++ b/spec/index.html
@@ -654,8 +654,8 @@
     Since all these implementations pre-date this 
     specification, some more interoperable choices include the following:
     
-    * to forbid cycles in YAML-LD documents
-    * to consider all references in YAML-LD as static, 
+    * forbidding cycles in YAML-LD documents
+    * considering all references in YAML-LD as static, 
       i.e., a shorthand way to repeat specific patterns
 
     </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -357,9 +357,9 @@
     <p>
       A YAML-LD stream is a YAML stream of YAML-LD documents.
       Note that each document in a stream is independent
-      from the others:
+      from the others;
       each one has its own context, YAML directives,
-      named anchors and so on.
+      named anchors, and so on.
     </p>
     <p>
       A YAML-LD document is a [[YAML]] document
--- a/spec/index.html
+++ b/spec/index.html
@@ -287,8 +287,8 @@
       To ease writing and collaborating on JSON-LD documents, it is a common practice
       to serialize them as YAML.
       This requires a registered media type, not only to enable content negotiation
-      of linked data documents in YAML, but even to define the expected behavior of
-      applications that processes these documents, including fragment identifiers and
+      of linked data documents in YAML, but also to define the expected behavior of
+      applications that process these documents, including fragment identifiers and
       interoperability considerations.
     </p>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -601,7 +601,7 @@
     givenName: John
     ```
     
-    transforming the above entry into a JSON-LD document
+    Transforming the above entry into a JSON-LD document
     results in:
     
     ```json
--- a/spec/index.html
+++ b/spec/index.html
@@ -221,7 +221,7 @@
         to represent information serialized as JSON,
         including Linked Data.
         This document defines how to serialize linked data
-        in YAML documents.
+        in YAML.
         Moreover, it registers the application/ld+yaml media type.
       </p>
     </section>
@@ -244,15 +244,12 @@
         Since YAML is more expressive than JSON,
         both in the available data types and in the document structure
         (see [[I-D.ietf-httpapi-yaml-mediatypes]]),
-        this document identifies constraints on YAML documents
-        such that they can be used to represent JSON-LD documents.
+        this document identifies constraints on YAML
+        such that it can be used to represent JSON-LD documents.
       </p>
     </section>
 
   <section id="conformance">
-    <p>A <a>YAML-LD document</a> complies with this specification if ...</p>
-    <p class="ednote">Define <dfn>YAML-LD document</dfn> somewhere.</p>
-
     <p>This specification makes use of the following namespace prefixes:</p>
     <table class="simple">
       <thead><tr>
@@ -301,6 +298,11 @@
         <li>YAML supports more data types than JSON.</li>
         <li>the structure of a YAML document &mdash; that is, a named YAML representation graph &mdash;
         is a rooted, directed graph that can have cycles.</li>
+        <li>YAML has the concept of stream, which is a sequence of documents.
+            While a stream usually contains one document,
+            streams with multiple documents are used to aggregate multiple,
+            related, documents into a single file or network stream.
+        </li>
 
     <p>
       The first goal of this specification is to allow a JSON-LD document to be
@@ -353,14 +355,22 @@
     <h2>Core Requirements</h2>
 
     <p>
-      A YAML-LD document is a [[YAML]] document that can be interpreted as Linked Data [[LINKED-DATA]].
+      A YAML-LD stream is a YAML stream of YAML-LD documents.
+      Note that each document in a stream is independent
+      from the others:
+      each one has its own context, YAML directives,
+      named anchors and so on.
+    </p>
+    <p>
+      A YAML-LD document is a [[YAML]] document
+      that can be interpreted as Linked Data [[LINKED-DATA]].
     </p>
     <p>
       It MUST be encoded in UTF-8, to ensure interoperability with [[JSON]]. 
     </p>
     <p>
       Comments in YAML-LD documents
-      are treated as white space
+      are treated as white space.
       This behavior is consistent with other
       Linked Data serializations like [[TURTLE]].
       See Interoperability considerations of [[I-D.ietf-httpapi-yaml-mediatypes]]
@@ -381,7 +391,7 @@
     </p>
     <p data-format="markdown">
       Example: The following YAML-LD document
-      contains alias nodes for the `{"@id": "country:ITA"}` object:
+      contains alias nodes for the `{"@id": "countries:ITA"}` object:
 
       ```yaml
       %YAML 1.2
--- a/spec/index.html
+++ b/spec/index.html
@@ -587,7 +587,7 @@
 
     #### Why does YAML-LD not preserve comments?
     <p class="ednote">
-    [[JSON]] (and hence [[JSON-LD]]) does not support comments,
+    [[JSON]] (and hence [[JSON-LD11]]) does not support comments,
     and other Linked Data serialization formats
     that support comments (such as [[TURTLE]])
     do not provide a means to preserve them


ioggstream:
TODO: format examples

spec/index.html
```diff
@@ -243,7 +243,7 @@ <h2>Introduction</h2>
       <p>
         Since YAML is more expressive than JSON,
         both in the available data types and in the document structure
-        (see I-D.ietf-yaml-mediatypes),
+        (see [[I-D.ietf-yaml-mediatypes]]),
```
gkellogg, line 246:
```suggestion
        (see [[I-D.ietf-httpapi-yaml-mediatypes]]),
```

gkellogg: What is defined in `localBiblio` is "I-D.ietf-httpapi-yaml-mediatypes". Of course, that could be changed, but we need to be sure it's done consistently throughout the document.

spec/index.html
```diff
@@ -283,9 +283,152 @@ <h2>Introduction</h2>
...
+      -
+        "@id": http://example.org/3
+        title: Example 3
+    ```
+
+
+    </p>
   </section>
+  <section id="specifications" class="normative">
+    <h2>Specifications</h2>
```
gkellogg, line 348:
```suggestion
    <h2>Core Requirements</h2>
```

gkellogg:
Thanks for taking a lead in this document @ioggstream.

spec/index.html
```diff
@@ -283,9 +283,152 @@ <h2>Introduction</h2>
...
+    <p>
+      The first goal of this specification is to allow to process a JSON-LD document
+      and serialize it in YAML, and then back to JSON-LD without losing any information.
+
+      This is always possible, because a YAML representation graph can always represent
+      a non-cyclic tree, because JSON data types are a subset of YAML's, and because
+      JSON encoding is UTF-8.
+    </p>
+
+    <p>Example: the JSON-LD document below
```
gkellogg, line 311:
```suggestion
    <p data-format="markdown>Example: the JSON-LD document below
```

spec/index.html
```diff
@@ -283,9 +283,152 @@ <h2>Introduction</h2>
...
+      and MAY not be persisted when interpreting the document as JSON-LD.
+
+    </p>
+    <p> 
+      A YAML-LD document MAY contain named anchors and alias nodes, 
+      but its representation graph MUST NOT contain cycles.
+      When interpreting the document as JSON-LD,
+      alias nodes MUST be resolved by value to their target nodes.
+    </p>
+    <p>
```
gkellogg, line 376:
```suggestion
    <p data-format="markdown>
```

gkellogg:
The `data-format="markdown"` bits are a hack, but this can be cleaned up and formalized later, perhaps using some tabbed examples, as in the JSON-LD 1.1 specs.

spec/index.html
ioggstream: correct!

ioggstream:
Thanks Gregg. I need to learn how to "re-spec", but I think we can focus on the ideas for now :)

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+    ```
+
+
+    </p>
   </section>
+  <section id="specifications" class="normative">
+    <h2>Core Requirements</h2>
+
+    <p>
+      A YAML-LD document is a [[YAML]] document that can be interpreted as linked data.
```
gkellogg, line 351:
```suggestion
      A YAML-LD document is a [[YAML]] document that can be interpreted as Linked Data [[LINKED-DATA]].
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+      See Interoperability considerations of [[I-D.ietf-httpapi-yaml-mediatypes]] for more details.
+    </p>
+    <p>
+      Since named anchors are a serialization detail,
+      such names
+      MUST NOT be used to convey relevant information,
+      MAY be altered when processing the document,
+      and MAY not be persisted when interpreting the document as JSON-LD.
+
+    </p>
```
gkellogg, line 374: Probably should provide a reference for "named anchors", clarify that they are part of an _expanded profile_, and that when transforming to the _internal representation_, references are expanded in place. This contrasts with your **MUST NOT** statement, but I would expect a YAML de-serializer to do the heavy lifting for this, while removing any notion of those anchors and references in the post-processed form.

gkellogg: Okay, much of what I said is described below, but  maybe the paragraph needs to be tweaked to allow for this.

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
       <h3>Examples</h3>
       <p class="ednote">FIXME</p>
     </section>
+    <section id="faq" class="informative" data-format="markdown">
+      **NOTE** REMOVE THIS SECTION BEFORE PUBLICATION.
+
+      <h3>FAQ</h3>
+
+      #### Why YAML-LD does not preserve comments?
```
gkellogg, line 577:
```suggestion
      #### Why does YAML-LD not preserve comments?
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
       <h3>Examples</h3>
       <p class="ednote">FIXME</p>
     </section>
+    <section id="faq" class="informative" data-format="markdown">
+      **NOTE** REMOVE THIS SECTION BEFORE PUBLICATION.
```
gkellogg, line 573:
```suggestion
      <p class="denote">REMOVE THIS SECTION BEFORE PUBLICATION.</p>
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+        "givenName": "John",
+        "http://www.w3.org/ns/json-ld/yaml-ld#comment": [
+          "First comment",
+          "Second comment"
+        ]
+      }
+      ```
+
+      
+      #### Why YAML-LD does not extend JSON-LD data structure ?
```
gkellogg, line 608:
```suggestion
      #### Why does YAML-LD not extend the JSON-LD data model ?
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+      <p class="ednote">
+      [[JSON]] represent simple trees while [[YAML]] can support
+      rooted, directed graphs with references and cycles.
+
+      The above structures cannot be preserved when serializing
+      to JSON-LD and - with respect to cycles - they serialization
```
gkellogg, line 614:
```suggestion
      to JSON-LD and - with respect to cycles - the serialization
```

spec/index.html
gkellogg: We also need to clarify that these features are **NOT** supported in the base profile, only in the extended profile.

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+      will fail.
+
+      Programming languages such as java an python already support
+      YAML representation graphs, but those implementation might behave
```
gkellogg, line 618:
```suggestion
      YAML representation graphs, but these implementations may behave
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
   <section id="basic-concepts" class="informative">
     <h2>Basic Concepts</h2>
 
-    <p>FIXME.</p>
+    <p>
+      To ease writing and collaborating on JSON-LD documents, it is a common practice
+      to serialize them as YAML.
+      This requires not only to register a media type to enable content negotiation
```
TallTed, line 289:
```suggestion
      This requires a registered media type, not only to enable content negotiation
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+      of linked data documents in YAML, but even to define the expected behavior of
+      applications that processes these documents, including fragment identifiers and
+      interoperability consideration.
```
TallTed, line 292:
```suggestion
      interoperability considerations.
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+      Since all these implementations pre-date this 
+      specification, the more interoperable choices would be:
+      
+      * to forbid cycles in YAML-LD documents;
+      * to consider all references as static, e.g.
+        a shorthand way to repeat specific patterns.
+
+      </p>
+    </section>
+
```
gkellogg, line 656: We should probably discuss how YAML dates and dateTime (equivalents) are treated. Presumably, as part of the extended profile, they can be expanded into their equivalent JSON-LD/XSD form, although considering the affect of context definitions for properties of which they are the value may complicate this somewhat.

Also, are YAML-LD processors required to parse all input regardless of the stated profile (expanded or basic), or only if the appropriate profile is set? I think the rule of profile parameters is that they don't change the semantics, so they either MUST process all forms, or error if the form is incorrect. (should be an issue).

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+      applications that processes these documents, including fragment identifiers and
+      interoperability consideration.
+
+      This is because YAML is more flexible than JSON:
+
+      * it supports different encodings, including UTF-8, UTF-16 and UTF-32;
+      * it supports more data types than JSON;
+      * the structure of its documents, that is named YAML representation graph,
+        is a tree that can have cycles.
+    </p>
```
TallTed, line 300:
```suggestion
    <p>
      This is because YAML is more flexible than JSON:
    </p>

      <ul>
        <li>YAML supports different encodings, including UTF-8, UTF-16, and UTF-32.</li>
        <li>YAML supports more data types than JSON.</li>
        <li>the structure of a YAML document &mdash; that is, a named YAML representation graph &mdash;
        is a tree that can have cycles.</li>
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+      of linked data documents in YAML, but even to define the expected behavior of
+      applications that processes these documents, including fragment identifiers and
+      interoperability consideration.
+
+      This is because YAML is more flexible than JSON:
+
+      * it supports different encodings, including UTF-8, UTF-16 and UTF-32;
+      * it supports more data types than JSON;
+      * the structure of its documents, that is named YAML representation graph,
+        is a tree that can have cycles.
```
pchampin, line 299: If it has cycles, then it is *not* a tree.

I think the correct notion here is that of [rooted graph](https://en.wikipedia.org/wiki/Rooted_graph)

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+    </p>
+
+    <p>
+      The first goal of this specification is to allow to process a JSON-LD document
+      and serialize it in YAML, and then back to JSON-LD without losing any information.
```
TallTed, line 304:
```suggestion
      The first goal of this specification is to allow a JSON-LD document to be
      processed and serialized into YAML, and then back into JSON-LD, without
      losing any information.
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+        {"@id": "http://example.org/1", "title": "Example 1"},
+        {"@id": "http://example.org/2", "title": "Example 2"},
+        {"@id": "http://example.org/3", "title": "Example 3"}
+      ]
+    }
+    ```
+
+    can be serialized as YAML as follows. Note that - since `@`
+    is a reserved character in YAML, entries containing it 
+    needs to be enclosed in double quotes.
```
TallTed, line 326:
```suggestion
    can be serialized as YAML as follows. Note that entries 
    containing `@` MUST be enclosed in double quotes because
    `@` is a reserved character in YAML.
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+    </p>
   </section>
+  <section id="specifications" class="normative">
+    <h2>Core Requirements</h2>
+
+    <p>
+      A YAML-LD document is a [[YAML]] document that can be interpreted as linked data.
+    </p>
+    <p>
+      It MUST be encoded in UTF-8, to improve interoperability with [[JSON]].
```
TallTed, line 354:
```suggestion
      It MUST be encoded in UTF-8, to ensure interoperability with [[JSON]]. 
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+    <p>
+      It MUST be encoded in UTF-8, to improve interoperability with [[JSON]]. 
+    </p>
+    <p>
+      Comments in YAML-LD documents
+      MAY be removed when interpreting the document
+      as JSON-LD.
+      It is then legitimate for a YAML-LD implementation to ignore comments,
+      and relevant information SHOULD be provided using specific predicates
+      such as rdfs:comment.
```
TallTed, line 362:
```suggestion
      It is therefore legitimate for a YAML-LD implementation to ignore comments,
      and relevant information SHOULD be provided using specific predicates
      such as `rdfs:comment`.
```

spec/index.html
TallTed, line 374:
```suggestion
      and MAY be dropped when interpreting the document as JSON-LD.
    </p>
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+    </p>
+    <p> 
+      A YAML-LD document MAY contain named anchors and alias nodes, 
+      but its representation graph MUST NOT contain cycles.
+      When interpreting the document as JSON-LD,
+      alias nodes MUST be resolved by value to their target nodes.
+    </p>
+    <p data-format="markdown">
+      Example: The following YAML-LD document
+      contains alias nodes for the {"@id": "country:ITA"} object;
```
TallTed, line 383:
```suggestion
      contains alias nodes for the `{"@id": "country:ITA"}` object:
```

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+      - "@id": http://people.example/Homer
+        name: Homer Simpson
+        nationality: *ITA
+      - "@id": http://people.example/Lisa
+        name: Lisa Simpson
+        nationality: *ITA
+      ```
+
+      While the representation graph (and eventually the in-memory representation
+      of the data structure, e.g. a python dictionary or a Java hashmap) will still
```
TallTed, line 403:
```suggestion
      of the data structure, e.g., a Python dictionary or a Java hashmap) will still
```

spec/index.html
pchampin, line 577: A parallel could also be made with other concrete syntaxes of RDF that *do* support comments (RDF/XML, Turtle).
I don't think that anybody ever suggested that these comments find their way into the graph...

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
+            "@id": "http://people.example/Lisa",
+            "full_name": "Lisa Simpson",
+            "country": {
+              "@id": "countries:ITA"
+            }
+          }
+        ]
+      }
+
+
```
TallTed, line 433:
```suggestion
      }
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+      <h3>FAQ</h3>
+
+      #### Why YAML-LD does not preserve comments?
+      <p class="ednote">
+      [[JSON]] does not support comments.
+      While YAML-LD could define a specific predicate for that,
+      that is not enough because for example
+      in JSON the order of keywords is not preserved would
+      displace the comments.
```
TallTed, line 583:
```suggestion
      While YAML-LD could define a specific predicate for comments,
      that is insufficient because, for example,
      the order of keywords is not preserved in JSON, so the
      comments could be displaced.
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+          "First comment",
+          "Second comment"
+        ]
+      }
+      ```
+
+      
+      #### Why YAML-LD does not extend JSON-LD data structure ?
+      <p class="ednote">
+      [[JSON]] represent simple trees while [[YAML]] can support
```
TallTed, line 610:
```suggestion
      [[JSON]] only represents simple trees while [[YAML]] can support
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+      rooted, directed graphs with references and cycles.
+
+      The above structures cannot be preserved when serializing
+      to JSON-LD and - with respect to cycles - they serialization
+      will fail.
+
+      Programming languages such as java an python already support
```
TallTed, line 617:
```suggestion
      Programming languages such as Java and Python already support
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+          value: *value
+          unit: degC
+      valve2:
+        temperature: *temp100C
+      ```
+
+      In python, processing this entry I get the following
+      structure, that preserves only the references to
+      mutable objects (e.g. the `temperature` dict)
+      but not the one to the `value` keyword (which is a scalar).
```
TallTed, line 636:
```suggestion
      Processing this entry in Python, I get the following
      structure that preserve the references to
      mutable objects (e.g., the `temperature` dict)
      but not to scalars (e.g., the `value` keyword).
```

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+        "valve2": { "temperature": temperature }
+      }
+      ```
+
+      Since all these implementations pre-date this 
+      specification, the more interoperable choices would be:
+      
+      * to forbid cycles in YAML-LD documents;
+      * to consider all references as static, e.g.
+        a shorthand way to repeat specific patterns.
```
TallTed, line 652:
```suggestion
      * to forbid cycles in YAML-LD documents
      * to consider all references in YAML-LD as static, 
        i.e., a shorthand way to repeat specific patterns
```

spec/index.html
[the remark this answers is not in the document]
gkellogg: Perhaps we can figure out some markup to add “rationale” notes, ideally with links to issues that record resolutions, either inline, or in an appendix.

spec/index.html
gkellogg: Yes, and reserializing these formats does loose comments, so it makes sense that YAML-LD comments would also be dropped. `rdfs:coment`, or similar, can be used in any serialization.

ioggstream:
@TallTed Thanks for the first batch of suggestions! I merged all the ones that did not require further discussion.

spec/index.html
ioggstream: @TallTed refactoring is ok, but we don't need the MUST at line 329 because we are just explaining how YAML works, not mandating something new.

Is  there an alternative to `needs to be enclosed`? Can we retain it?

spec/index.html
ioggstream: @gkellogg I am opening a separate issue for discussing profiles: https://github.com/json-ld/yaml-ld/issues/35

My proposal for now is to retain the following information, and move them around when fixing #35 

WDYT?

spec/index.html
```diff
@@ -421,6 +569,91 @@ <h2>Fragment identifiers</h2>
...
+      temperature = { "value": 100, "unit": "degC" }
+      document = {
+        "value": 100,
+        "valve1": { "temperature": temperature },
+        "valve2": { "temperature": temperature }
+      }
+      ```
+
+      Since all these implementations pre-date this 
+      specification, the more interoperable choices would be:
```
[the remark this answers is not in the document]
ioggstream: @TallTed in this FAQ I just want to provide information that are useful for the discussion without stating anything.

Is  "would be" -> "might be" ok?

gkellogg:
Looks like Encoding Considerations use of `data-cite` needs to end with a  fragment identifier, or use the `[[YAML media type]]` notation.

`See <a data-cite="I-D.ietf-httpapi-yaml-mediatypes#">YAML media type</a>.`

gkellogg:
Note that `data-cite` is no longer recommended by ReSpec, and they suggest `xref` instead, but it doesn't seem to actually work with specs we reference.

spec/index.html
ioggstream: IIUC YAML 1.2 does not have a native date type anymore,  see https://github.com/yaml/yaml-spec/issues/268

Should be discussed after merging imho https://github.com/json-ld/yaml-ld/issues/35

spec/index.html
```diff
@@ -283,9 +283,158 @@ <h2>Introduction</h2>
...
+        <li>the structure of a YAML document &mdash; that is, a named YAML representation graph &mdash;
+        is a tree that can have cycles.</li>
+
+    <p>
+      The first goal of this specification is to allow a JSON-LD document to be
+      processed and serialized into YAML, and then back into JSON-LD, without
+      losing any information.
+
+      This is always possible, because a YAML representation graph can always represent
+      a non-cyclic tree, because JSON data types are a subset of YAML's, and because
```
ioggstream, line 311:
```suggestion
        is a rooted, directed graph that can have cycles.</li>

    <p>
      The first goal of this specification is to allow a JSON-LD document to be
      processed and serialized into YAML, and then back into JSON-LD, without
      losing any information.

      This is always possible, because a YAML representation graph can always represent
      a tree, because JSON data types are a subset of YAML's, and because
```

spec/index.html
ioggstream: @pchampin replaced. Could you PTAL and check if there are other incorrect terms, please? Thx++ R.

spec/index.html
ioggstream: Ok, I'll replace the FAQ above with some Turtle example. That's neat and clear.

spec/index.html
gkellogg: > IIUC YAML 1.2 does not have a native date type anymore, see [yaml/yaml-spec#268](https://github.com/yaml/yaml-spec/issues/268)

Just as well. But, it may come up again in the context of CBOR.

spec/index.html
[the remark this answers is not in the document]
ioggstream: @pchampin I suggest to try to put everything that is relevant in the spec.

FAQ are useful during the editing process for people that want to comment. Once a document is released,
I think that most of the FAQ can just fit a blog post on the WG repository.

In general, most of the rationale will be on the repository, and we probably shouldn't stuff all the issue discussion inside the spec.

spec/index.html
ioggstream: @gkellogg if that's ok, just resolve this conversation. This comment should be already linked in #35

spec/index.html
ioggstream:
```suggestion
    can be serialized as YAML as follows.
    Note that entries 
    starting with  `@` need to be enclosed in double quotes or escaped because
    `@` is a reserved character in YAML.
```

spec/index.html
ioggstream: I replaced both FAQ and the text. It's simpler now. 84b05759a55e789819315764fd5b895531fbd403

ioggstream:
> Note that `data-cite` is no longer recommended by ReSpec, and they suggest `xref` instead, but it doesn't seem to actually work with specs we reference.

Filed #40

spec/index.html
```diff
@@ -421,6 +569,97 @@ <h2>Fragment identifiers</h2>
       <h3>Examples</h3>
       <p class="ednote">FIXME</p>
     </section>
+    <section id="faq" class="informative" data-format="markdown">
+      <p class="ednote">REMOVE THIS SECTION BEFORE PUBLICATION.</p>
```
gkellogg, line 573: This section should pop out to be Appendix B IMO. Otherwise, it seems to be part of IANA Considerations.

spec/index.html
```diff
@@ -283,9 +283,157 @@ <h2>Introduction</h2>
...
 
+    <p>
+      A YAML-LD document is a [[YAML]] document that can be interpreted as Linked Data [[LINKED-DATA]].
+    </p>
+    <p>
+      It MUST be encoded in UTF-8, to ensure interoperability with [[JSON]]. 
+    </p>
+    <p>
+      Comments in YAML-LD documents
+      MAY be removed when interpreting the document
```
gkellogg, line 363: The MAY wording may be odd, since it implies that there is process that is optional. If so, what is the process for implementations wanting to preserve comments? And how could they possibly be round-tripped? It may be better to simply require that they be dropped.

gkellogg (approved):
In general, this is great; just a couple of small nits.

spec/index.html
ioggstream: > ... the process is optional

I think that it is optional. One could e.g.

```
g = rdflib.Graph()
g.parse("foo.yamlld")
g.serialize("bar.yamlld")  #  retain comments
g.serialize("bar.ttl")         # retain comments ?
g.serialize("bar.jsonld")  # strip comment
```

> what is the process for ... preserving comments?

Maybe there's somebody out there that can do better than me, so I don't want to limit him.

> require that they be dropped

An implementation that is capable of preserving comments e.g. in its internal representation, or when converting YAML to Turtle would thus be non conformant.

gkellogg: No other RDF serialization format that I'm aware of retains comments post-parsing. As with [Turtle](https://www.w3.org/TR/turtle/#sec-grammar-comments), we should probably just treat comments as white space. Normative behavior on ignoring them isn't necessary.

Are there _any_ examples of other serialization formats (RDF anyway) retaining comments? Isn't that the purpose of something like `rdfs:comment`, which specifically allows for a way of introducing non-syntactic comments in a document?

ioggstream: ruamel.yaml is a YAML implementation that preserves comments, so I just don't want forbid something that's actually feasible, e.g.

```
# pip install ruamel.yaml
from ruamel.yaml import YAML
from sys import stdout

yaml = YAML()
data = yaml.load("""
# Ciao
a: 1
""")
yaml.dump(data, stdout)
```
I think that "treating as white space" is ok, without normative behavior.

ioggstream:
@gkellogg all set. Feel free to merge, now.

spec/index.html
```diff
@@ -652,7 +652,7 @@ <h3>FAQ</h3>
     ```
 
     Since all these implementations pre-date this 
-    specification, the more interoperable choices would be:
+    specification, some more interoperable choices include the following:
     
     * to forbid cycles in YAML-LD documents
     * to consider all references in YAML-LD as static,
```
TallTed, line 658:
```suggestion
    * to consider all references in YAML-LD as static
```

ioggstream:
Rebased.

spec/index.html
```diff
@@ -283,9 +283,156 @@ <h2>Introduction</h2>
   <section id="basic-concepts" class="informative">
     <h2>Basic Concepts</h2>
 
-    <p>FIXME.</p>
+    <p>
+      To ease writing and collaborating on JSON-LD documents, it is a common practice
+      to serialize them as YAML.
+      This requires a registered media type, not only to enable content negotiation
+      of linked data documents in YAML, but even to define the expected behavior of
+      applications that processes these documents, including fragment identifiers and
```
TallTed, line 291:
```suggestion
      of linked data documents in YAML, but also to define the expected behavior of
      applications that process these documents, including fragment identifiers and
```

spec/index.html
```diff
@@ -423,5 +570,96 @@ <h3>Examples</h3>
...
+      "valve1": { "temperature": temperature },
+      "valve2": { "temperature": temperature }
+    }
+    ```
+
+    Since all these implementations pre-date this 
+    specification, some more interoperable choices include the following:
+    
+    * to forbid cycles in YAML-LD documents
+    * to consider all references in YAML-LD as static,
```
TallTed, line 658:
```suggestion
    * forbidding cycles in YAML-LD documents
    * considering all references in YAML-LD as static, 
```

anatoly-scherbakov (approved):

spec/index.html
```diff
@@ -353,14 +355,22 @@ <h2>Basic Concepts</h2>
     <h2>Core Requirements</h2>
 
     <p>
-      A YAML-LD document is a [[YAML]] document that can be interpreted as Linked Data [[LINKED-DATA]].
+      A YAML-LD stream is a YAML stream of YAML-LD documents.
+      Note that each document in a stream is independent
+      from the others:
+      each one has its own context, YAML directives,
+      named anchors and so on.
```
TallTed, line 362:
```suggestion
      from the others;
      each one has its own context, YAML directives,
      named anchors, and so on.
```

spec/index.html
```diff
@@ -353,14 +355,22 @@ <h2>Basic Concepts</h2>
...
+    </p>
+    <p>
+      A YAML-LD document is a [[YAML]] document
```
TallTed, line 365: For consistency with earlier content, there should be no "YAML document", but I wonder whether that means that there should also be no YAML-LD document -- i.e., whether YAML-LD must be streams just like YAML itself.
```suggestion
      A YAML-LD document is [[YAML]]
```

gkellogg:

spec/index.html
gkellogg: I'm leaving this suggestion unmerged based on @ioggstream's comments on today's meeting. It can be re-added as an issue or PR if it is inconsistent.

Issue #38, opened 2022-07-03
Conformance section


ioggstream:
## I expect

in conformance section

-  [ ] include yaml 
- [ ] include HTTP
- [ ] include linked data

PR #39, opened 2022-07-03
Fix: #38. Stubs Conformance.

ioggstream:
## This PR

- stubs conformance

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/39.html" title="Last updated on Jul 6, 2022, 9:19 PM UTC (fb3dba5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/39/971051f...fb3dba5.html" title="Last updated on Jul 6, 2022, 9:19 PM UTC (fb3dba5)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -260,6 +260,7 @@
     </p>
     <p>
       The terms YAML, "YAML document", "YAML representation graph",
+      "YAML stream", "YAML directive",
       "node", "scalar",
       "named anchor" and "alias nodes" are imported from [[YAML]].
     </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -256,7 +256,7 @@
       The term media type is imported from [[RFC6838]].
     </p>
     <p>
-      The terms JSON, "JSON document", "JSON string" are imported from [[JSON]].
+      The terms JSON, "JSON document", and "JSON string" are imported from [[JSON]].
     </p>
     <p>
       The terms YAML, "YAML document", "YAML representation graph",
--- a/spec/index.html
+++ b/spec/index.html
@@ -262,7 +262,7 @@
       The terms YAML, "YAML document", "YAML representation graph",
       "YAML stream", "YAML directive",
       "node", "scalar",
-      "named anchor" and "alias nodes" are imported from [[YAML]].
+      "named anchor", and "alias nodes" are imported from [[YAML]].
     </p>
     <p>
       The term "content negotiation" is imported from [[RFC9110]].
--- a/spec/index.html
+++ b/spec/index.html
@@ -252,7 +252,26 @@
   <section id="conformance">
     <p>A <a>YAML-LD document</a> complies with this specification if ...</p>
     <p class="ednote">Define <dfn>YAML-LD document</dfn> somewhere.</p>
-
+    <p>
+      The term media type is imported from [[RFC6838]].
+    </p>
+    <p>
+      The terms JSON, "JSON document", "JSON string" are imported from [[JSON]].
+    </p>
+    <p>
+      The terms YAML, "YAML document", "YAML representation graph",
+      "node", "scalar",
+      "named anchor" and "alias nodes" are imported from [[YAML]].
+    </p>
+    <p>
+      The term "content negotiation" is imported from [[RFC9110]].
+    </p>
+    <p>
+      The terms "fragment" and "fragment identifier" in this document are to be interpreted as in [[URI]].
+    </p>
+    <p>
+      The term "Linked Data" is imported from [[JSON-LD]].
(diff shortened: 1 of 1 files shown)


gkellogg (commented):
Might want to make actual definition references using `data-cite` for these terms. The JSON-LD spec created a "Definitions" section and defined the imported terms locally, which were themselves references. Might want to do something similar for consistency.

E.g.:

```
  <dt><dfn data-cite="JSON-LD11#dfn-json-ld-document">JSON-LD document</dfn></dt><dd>
    A <a>JSON-LD document</a> is a serialization of
    an <a>RDF dataset</a>.
    See the <a data-cite="JSON-LD11#json-ld-grammar">JSON-LD Grammar</a> section in JSON-LD 1.1 for a formal description.
  </dd>
```

These are defined in https://github.com/w3c/json-ld-wg/blob/main/common/terms.html and used with some clever code which tries to only define those terms actually used. I'm not proposing that we do that here, or that we use the local definition for for imported terms, necessarily, but structure them similarly.

ioggstream:
I suggest to do this referencing job in a subsequent editorial PR once the referenced terms are consolidated
filing another issue like  #40

> JSON-LD ... defined the imported terms locally

IIUC, in json-ld there are copy-pasted definitions from other documents. Is that correct?
I think we should just use references, so that people is encouraged to read the original documents:
an implementer is expected to read the JSON, JSON-LD and YAML spec.
The fact that they won't do it should be addressed e.g. in the Appendix.

In my recent ramblings on other specs I spent a good amount of time replacing copied definitions with hrefs to make them coherent with the latest HTTP spec (RFC9110): this work would have been easier if they just used hrefs :)


WDYT?

anatoly-scherbakov (approved):

gkellogg:
> IIUC, in json-ld there are copy-pasted definitions from other documents. Is that correct? I think we should just use references, so that people is encouraged to read the original documents: an implementer is expected to read the JSON, JSON-LD and YAML spec. The fact that they won't do it should be addressed e.g. in the Appendix.

Not exactly, consider the term "JSON object". It has an entry in the _dl_ for [Terms imported from Other Specifications](https://www.w3.org/TR/json-ld11/#terms-imported-from-other-specifications), but the key is a reference to [RFC8259](https://datatracker.ietf.org/doc/html/rfc8259#section-4). But, by defining a _dfn_ for this, we can simply use `<a>JSON Object</a>` rather than the more lengthly `<a data-cite="RFC8259#section-4">JSON Object</a>` within the body of the document for every occurrence; it will use the reference directly into RFC8259, rather than that in the _dl_. It simplifies later usage, and provides a single place for us to describe the terms we are importing, vs defining. Yes, there is a brief description, but that is mostly to describe the context for usage within JSON-LD, not to repeat the definition from the referenced spec.

ioggstream:
> brief description ...

@gkellogg ah, ok. I'll file it in a separate, editorial, issue. Since I'm working on 4 other I-Ds, it would be great if during today's call we could find someone volounteering on those editorial stuff.

ioggstream:
Moved the other issue to #47. @gkellogg feel free to merge

spec/index.html
```diff
@@ -252,7 +252,27 @@ <h2>Introduction</h2>
...
+      The term media type is imported from [[RFC6838]].
+    </p>
+    <p>
+      The terms JSON, "JSON document", "JSON string" are imported from [[JSON]].
+    </p>
+    <p>
+      The terms YAML, "YAML document", "YAML representation graph",
+      "YAML stream", "YAML directive",
+      "node", "scalar",
+      "named anchor" and "alias nodes" are imported from [[YAML]].
```
TallTed, line 265:
```suggestion
      "named anchor", and "alias nodes" are imported from [[YAML]].
```

gkellogg (approved):

anatoly-scherbakov (approved):

spec/index.html
```diff
@@ -262,7 +262,7 @@ <h2>Introduction</h2>
       The terms YAML, "YAML document", "YAML representation graph",
```
TallTed, line 262: "YAML document" has been carefully edited out of most of this document, so perhaps it should be removed here as well? 
```suggestion
      The terms YAML, "YAML representation graph",
```

OR13 (approved):

gkellogg:

spec/index.html
ioggstream: The original text is ok to me. Let me just rebase the PR @gkellogg :)

gkellogg: PR is merged, thanks.

Issue #7, opened 2022-05-27
File signature


nichtich:
As a data consumer
I want an indicator to tell me that a file is probably YAML-LD
So that I know when to expect YAML-LD

Strict checking whether a YAML document is valid YAML-LD requires to follow the full specification. Nevertheless
some kind of [magic file number](https://en.wikipedia.org/wiki/List_of_file_signatures) would be useful. [As suggested here](https://github.com/json-ld/yaml-ld/issues/6#issuecomment-1138301226) a YAML global tag could be used for this purpose (see [RFC 4151](https://datatracker.ietf.org/doc/html/rfc4151)): 

~~~yaml
!<tag:json-ld.org,2022>
$context: http://schema.org/
$type: Person
name: Pierre-Antoine Champin
~~~

YAML processors will raise a "unknown tag" error when trying to process the document without knowledge of YAML-LD. It can still be parsed as valid YAML but there is no default mapping to JSON. This is not a bug, but a feature.

pchampin:
>  I want an indicator to tell me that a file is probably YAML-LD
Assuming a [specific content-type](https://github.com/ietf-wg-httpapi/mediatypes/issues/8) gets registered, would this answer your issue?
Or do you want the indicator to be *in* the content itself?

ioggstream:
@nichtich some comments:

1. If we want to rely on YAML, we should consider that files could be normalized like that:

```
%YAML 1.2
---
!<tag:json-ld.org,2022>
$context: http://schema.org/
$type: Person
name: Pierre-Antoine Champin
```

This means that we cannot rely on magic numbers / file signatures.

2. I am reasoning on tags...

nichtich:
> we should consider that files could be normalized like that: [...]

There are several more ways to push down the start of actual content in a YAML file. YAML syntax is a beast.

> Or do you want the indicator to be in the content itself?

Text based file formats rarely have traditional fixed positon magic file numbers but applications should be able to scan the first lines of a YAML file to detect whether it's meant to be YAML-LD.

By the way another solution to this requirement is to state that YAML-LD document must include [$schema](https://json-schema.org/draft/2020-12/json-schema-core.html#keyword-schema) as its first key.

ioggstream:
> YAML syntax is a beast

This is not going to change :) 

> YAML-LD document must include [$schema](https://json-schema.org/draft/2020-12/json-schema-core.html#keyword-schema) as its first key

[...]

PR #50, opened 2022-07-05
Implement Best Practices section

anatoly-scherbakov:
- Update document boilerplate.
- Use case for issue #5.
- Use case for issue #6.
- Use Markdown for the first use case.
- Stub introduction / abstract
- Apply suggestions from code review
- Stub for security, interoperability and fragment identifier considerations.
- Apply suggestions from code review
- Adds a preliminary IANA section with placeholders.
- Apply suggestions from @ioggstream.
- Cleanup IANA and move Fragment Identifiers into the media type section.
- Move fragment identifiers section into IANA considerations.
- Update from code review.
- Fix: #7. Same as YAML.
- Implement `Makefile` and ignore IntelliJ files
- Write draft of the Best Practices chapter
- A few textual adjustments

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -443,52 +443,51 @@
     <section id="best-practices" class="informative">
       <h2>Best Practices</h2>
 
-      <p>In this section, we list a few recommendations to YAML-LD users which, although optional, might contain one or more useful thoughts.</p>
+      <p>Here, we propose to YAML-LD users a bit of advice which, although optional, might suggest one or two useful thoughts.</p>
 
        <div class="practice">
           <p class="practicedesc">
             <span id="use-json-ld-best-practices" class="practicelab">Follow JSON-LD best practices</span>
-            …in order to achieve a greater level of reusability, performance, and human friendliness of YAML-LD aware systems, follow [[json-ld-bp]] document that is as applicable to YAML-LD as it is to [[JSON-LD]].
+            …in order to achieve a greater level of reusability, performance, and human friendliness of YAML-LD aware systems. The [[json-ld-bp]] document is relevant to YAML-LD no less than it is to [[JSON-LD]].
           </p>
        </div>
 
       <div class="practice">
         <p class="practicedesc">
-          <span id="conceal-context" class="practicelab">Conceal JSON-LD Context if possible</span>
+          <span id="prebuilt-contexts" class="practicelab">Do not force users to author contexts</span>
 
-          A YAML-LD aware system which consumes YAML-LD data should not require the document author to specify an explicit context if the system is capable of inferring the proper context itself.
+          Instead, provide pre-built contexts that the user can reference by URL for majority of common use cases.
         </p>
       </div>
 
-      <p>YAML-LD is intended to simplify authoring of Linked Data for a wide range of domain experts: its target audience is not comprised solely of IT professionals. YAML is chosen as a medium to minimize syntactic noise, keep the authored documents concise and clear.</p>
+      <p>YAML-LD is intended to simplify authoring of Linked Data for a wide range of domain experts: its target audience is not comprised solely of IT professionals. YAML is chosen as a medium to minimize syntactic noise, to keep the authored documents concise and clear. [[JSON-LD]] (and hence YAML-LD) Context comprises a special language of its own. A requirement to <i>author</i> such a context would make the domain expert's job much harder, — which we, as system architects and developers, should try to avoid.</p>
 
-      <p>To that end, it is recommended that YAML-LD aware applications do not require the users to specify contexts.</p>
+      <div class="practice">
+        <p class="practicedesc">
+          <span id="conceal-contexts" class="practicelab">Use a default context</span>
+        </p>
 
-      <ul>
-        <li>JSON-LD (and hence YAML-LD) Context comprises a special language of its own. A requirement to author such a context would make the job of the domain expert much more difficult, which we usually are trying to avoid;</li>
-        <li>Alternatively, if the user has to specify the URL of the context, for them it will become a technical incantation of unclear meaning which they will have to copy-paste from one document to another.</li>
-      </ul>
+        If most, or all, of a user's documents are based on one particular context, try to make it the default in order to rescue the user from copy-pasting the same technical incantation from one document to another.
+      </div>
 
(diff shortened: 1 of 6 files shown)


spec/index.html
```diff
@@ -423,5 +440,80 @@ <h3>Examples</h3>
...
+            <span id="convenience-context" class="practicelab">Use YAML-LD Convenience Context</span>
+            <p>YAML-LD users may use a JSON-LD context provided as part of this specification, henceforth known as the <dfn>convenience
+          context</dfn>, which defines a standardized mapping of every `@`-keyword to a `$`-keyword, except `@context`.</p>
+         </p>
+       </div>
+
+      <aside class="example" data-transform="updateExample" title="Convenience Context">
+        <p>Code of the <a>convenience context</a>. Available as: <a href="https://yaml-ld.dev/context">https://yaml-ld.dev/context</a>.
+        </p>
+        <pre class="json" data-include="code/context.json" data-include-format="text"></pre>
```
gkellogg, line 505: It might be worth inlining these to avoid having included files dangle. Also, PR preview doesn't seem to handle these well.

spec/index.html
```diff
@@ -423,5 +440,80 @@ <h3>Examples</h3>
...
+
+      <div class="practice">
+        <p class="practicedesc">
+          <span id="conceal-contexts" class="practicelab">Use a default context</span>
+        </p>
+
+        If most, or all, of a user's documents are based on one particular context, try to make it the default in order to rescue the user from copy-pasting the same technical incantation from one document to another.
+      </div>
+
+      <p>For instance, according to [[json-ld-api]], `expand()` function of a JSON-LD processor accepts an `expandContext` argument which can be used to provide a default system context.</p>
```
gkellogg, line 473:
```suggestion
      <p>For instance, according to [[json-ld-api]], the `expand()` method of a JSON-LD processor accepts an `expandContext` argument which can be used to provide a default system context.</p>
```

gkellogg (commented):
Please consider wrapping lines on natural boundaries, which make it easier to make change suggestions through GitHub.

Not sure I agree with all of the best practices you propose, but it's an interesting point of discussion.

spec/index.html
anatoly-scherbakov: Indeed… I hoped that including files would help editing experience, to provide syntax highlighting in editors and to be able to perhaps autotest the included files. What is your view on these potential advantages?

anatoly-scherbakov:
Line wraps added,  thank you for pointing out.

spec/index.html
gkellogg: We did all JSON-LD examples inline, using some helper JS to avoid escaping. I implemented a CI mechanism to validate all the examples, and we should do something like that here as well; I can certainly help facilitate this if you can stand some Ruby :).

ReSpec has had some issues with referenced files that have bitten me before when using the spec-prod GH Action, but it is a fair thing to use. Too bad the PR-preview rendering seems broken.

spec/index.html
```diff
@@ -423,5 +440,98 @@ <h3>Examples</h3>
...
+
+          Instead, provide pre-built contexts that the user can reference by URL for majority of common use cases.
+        </p>
+      </div>
+
+      <p>YAML-LD is intended to simplify authoring of Linked Data for a wide range of domain experts: its target
+        audience is not comprised solely of IT professionals. YAML is chosen as a medium to minimize syntactic noise,
+        to keep the authored documents concise and clear. [[JSON-LD]] (and hence YAML-LD) Context comprises a special
+        language of its own. A requirement to <i>author</i> such a context would make the domain expert's job much
+        harder, — which we, as system architects and developers, should try to avoid.</p>
```
TallTed, line 469:
```suggestion
      <p>YAML-LD is intended to simplify the authoring of Linked Data for a wide range of domain experts; its target
        audience is not comprised solely of IT professionals. YAML is chosen as a medium to minimize syntactic noise,
        and to keep the authored documents concise and clear. [[JSON-LD]] (and hence YAML-LD) Context comprises a special
        language of its own. A requirement to <i>author</i> such a context would make the domain expert's job much
        harder &mdash; which we, as system architects and developers, should try to avoid.</p>
```

spec/index.html
```diff
@@ -423,5 +440,98 @@ <h3>Examples</h3>
...
+      <div class="practice">
+        <p class="practicedesc">
+          <span id="alias-keywords" class="practicelab">Alias JSON-LD keywords</span>
+
+          If possible, map JSON-LD keywords containing the `@` character to keywords that do not contain it.
+        </p>
+      </div>
+
+      <p>The `@` character is reserved as per [[YAML]] specification and thus requires quoting, as in the following
+        example:</p>
```
TallTed, line 492:
```suggestion
      <p>The `@` character is reserved in [[YAML]], and thus requires quoting (or escaping), as in the following
        example:</p>
```

spec/index.html
```diff
@@ -423,5 +440,98 @@ <h3>Examples</h3>
...
+        <p>Code of the <a>convenience context</a>. Available as:
+          <a href="https://yaml-ld.dev/context">https://yaml-ld.dev/context</a>.
+        </p>
+        <pre class="json" data-include="code/context.json" data-include-format="text"></pre>
+      </aside>
+
+      <p>The convenience context contains an alias to every JSON-LD keyword which can be aliased as per JSON-LD 1.1
+        specification, — which means all the keywords except <code>@context</code>. The reserved `@` character is
+        replaced by `$`, which is not reserved and therefore does not require quoting. Consider
+        <strong>Example 1</strong> reformatted using the convenience context:</p>
```
TallTed, line 527:
```suggestion
      <p>The convenience context contains an alias to every JSON-LD keyword which the JSON-LD 1.1
        specification permits aliasing &mdash; which means all the keywords except <code>@context</code>. The reserved `@` character is
        replaced by `$`, which is not reserved and therefore does not require quoting. Consider
        <strong>Example 1</strong> reformatted using the convenience context:</p>
```

VladimirAlexiev:
@anatoly-scherbakov 
> proposed hard-wired $ ←→ @ conversion is not feasible

Agreed it cannot be the only or the default option.

But I think there should be an option `sigil:` or `prefix_char:` where I can pick my own: `$` or `_` etc

gkellogg:

PR #56, opened 2022-07-10
Fix example: no escaping, only quote

ioggstream:
## This PR

Fixes example.

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/56.html" title="Last updated on Jul 10, 2022, 3:17 PM UTC (b5cd5d2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/56/567fda6...b5cd5d2.html" title="Last updated on Jul 10, 2022, 3:17 PM UTC (b5cd5d2)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -378,7 +378,7 @@
     <p>
       Can be serialized as YAML as follows.
       Note that entries
-      starting with  `@` need to be enclosed in quotes or escaped because
+      starting with  `@` need to be enclosed in quotes because
       `@` is a reserved character in YAML.
     </p>
 
@@ -389,13 +389,13 @@
       <!--
       %YAML 1.2
       ---
-      "@context": http://example.org/context.jsonld
-      \@graph:
+      '@context': http://example.org/context.jsonld
+      '@graph':
         -
-          "@id": http://example.org/1
+          '@id': http://example.org/1
           title: Example 1
         -
-          \@id: http://example.org/2
+          '@id': http://example.org/2
           title: Example 2
         -
           '@id': http://example.org/3


gkellogg (approved):

spec/index.html
```diff
@@ -378,7 +378,7 @@ <h2>Basic Concepts</h2>
     <p>
       Can be serialized as YAML as follows.
       Note that entries
-      starting with  `@` need to be enclosed in quotes or escaped because
+      starting with  `@` need to be enclosed in quotes because
       `@` is a reserved character in YAML.
```
TallTed, line 382: My read of [the YAML spec](https://yaml.org/spec/1.2.2/) with some reference to external docs like [Getting started with yaml, § Escaping Characters](https://riptutorial.com/yaml#escaping-characters) indicates that escaping *is* an option, though it's not being used here.

I believe it appropriate to mention that option in this side-note on this example, even though escaping is not being used here, because it can otherwise appear that we're saying that escaping is not an option.

A possible rephrasing of the note --

```suggestion
      Note that entries starting with `@` need to be
      enclosed in quotes (as shown in this example),
      or the `@` needs to be escaped,
      because `@` is a reserved character in YAML.
```

ioggstream: @TallTed see YAML §5.7

> escape sequences are only interpreted in [double-quoted scalars](https://yaml.org/spec/1.2.2/#double-quoted-style).

In particular my preliminary tests with pyyaml e.g.

```
\u0040context: http://foo.bar
```

result in an error because it's equivalent to 

```
@context: http://foo.bar
```

I think we can write a testlist and ask to YAML folks for details.

TallTed: Ah, I think I see.  Looks like I didn't read deeply enough.  YAML has a longer learning curve than I anticipated. :-/

I think there are some other mentions of escaping in the docs which may also need to be eliminated or edited carefully by someone more deeply versed in YAML subtleties than I.

TallTed: @ioggstream —
> In particular my preliminary tests with pyyaml e.g.
>
>  ```\u0040context: http://foo.bar```

I wonder if that didn't fail because `\u0040` is an _encoding_ of `@`, rather than an _escape_, which I think would be `\@` (or *possibly* `\\\u0040` [i.e., double-backslash to escape the leading backslash, then another backslash to start the unicode encoding -- in other words, equivalent to `\@`]).

PR #64, opened 2022-07-25
Update YAML reference to 1.2.2, addressing references to "YAML stream".

### LINKED ISSUE #61: Links to sections of YAML spec do not work
At PR #59, @gkellogg had introduced links to sections of the YAML specification which define particular terms of the spec.

## Why?

`[[YAML]]` is dereferenced by ReSpec as per [SpecRef](https://www.specref.org/?q=yaml) into http://yaml.org/spec/1.2/spec.html. Thus, `YAML#92-streams` is converted into http://yaml.org/spec/1.2/spec.html#92-streams.

Which does not work as expected, instead it redirects to https://yaml.org/spec/1.2.2/ and the fragment identifier is lost.

## What?

I would like to propose to change the YAML spec link to version 1.2.2 instead of 1.2. That should alleviate the redirect.

gkellogg:
Note that the title was changed to "YAML Ain’t Markup Language (YAML™) version 1.2.2" from "YAML Ain’t Markup Language (YAML™) version 1.2". Arguably, we should be using `[[YAML122]]` to be more explicit, and should probably note something about 1.2.2 being the minimum version supported.

Fixes #61.

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/64.html" title="Last updated on Jul 26, 2022, 7:53 PM UTC (e01af8f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/64/78a329e...e01af8f.html" title="Last updated on Jul 26, 2022, 7:53 PM UTC (e01af8f)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -297,6 +297,8 @@
         Each document in a stream has its own context, [[YAML]] directives, named anchors, and so on.
     </p>
 
+    <p class="note"><a>YAML streams</a> were introduced in YAML 1.2.2, and cannot be used in earlier versions.</p>
+    
     <p>A <dfn>YAML-LD document</dfn> is any <a>YAML document</a> from
       which a conversion to [[JSON]] produces
       a valid <a>JSON-LD document</a> which can be interpreted as [[LINKED-DATA]].
@@ -420,6 +422,7 @@
           While a stream often contains one document,
           streams with multiple documents are used to aggregate multiple,
           related, documents into a single file or network stream.
+          <div class="note"><a>YAML streams</a> were introduced in YAML 1.2.2, and cannot be used in earlier versions.</div>
       </li>
     </ul>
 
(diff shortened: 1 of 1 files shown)


anatoly-scherbakov (approved):

TallTed:
Yes, I think this does mean some additional edits are needed of the part where the doc says "YAML-LD is not bound to any specific version of YAML, though this doc is based on 1.2.2", which implies that earlier as well as newer versions of YAML can be used for YAML-LD.

gkellogg:
> Yes, I think this does mean some additional edits are needed of the part where the doc says "YAML-LD is not bound to any specific version of YAML, though this doc is based on 1.2.2", which implies that earlier as well as newer versions of YAML can be used for YAML-LD.

I'm not sure that updating the YAML reference affects this. It may mean that the concept of "YAML streams" is not appropriate if using an earlier version of the spec, but that doesn't really change the way people would interpret this document if using an older version of YAML.

TallTed:
Ah, I see... So the need is more for bit of prose along the lines of "YAML streams were introduced in YAML 1.2.2, and cannot be used in earlier versions" in the relevant section(s), more than a global.

anatoly-scherbakov (approved):

PR #65, opened 2022-07-25
Description of converting YAML to the internal representation.

### LINKED ISSUE #62: add section outlining conversion to Internal Representation
The primary issue is that there seem to be no normative specification for how to turn YAML into JSON, or more importantly, the internal representation shared between YAML, JSON, and other systems. If one can be found, it can be referenced, otherwise, this section would describe this transformation.

due 27 Jul 2022 (@gkellogg)

gkellogg:
Fixes #62.

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/65.html" title="Last updated on Aug 1, 2022, 10:27 PM UTC (a237088)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/65/9aa0d85...a237088.html" title="Last updated on Aug 1, 2022, 10:27 PM UTC (a237088)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -864,6 +864,12 @@
         If none exist, the document is invalid,
         and processing MUST end in failure.
       </p>
+
+      <p class="note">An invalid YAML document may have
+        an <a>alias node</a> used within the
+        <a>node anchor</a> definition which it references.
+        In some implemenentations, this may lead to a processing error.
+      </p>
     </section>
 
     <section id="convert-scalar">
--- a/spec/index.html
+++ b/spec/index.html
@@ -748,7 +748,7 @@
       </p>
 
       <ol>
-        <li>Create an empty <a>map</a> <var>named nodes</var>,
+        <li>Create an empty <a>map</a> <var>named nodes dictionary</var>,
           used the associate an anchor name with the <a>node</a>
           on which it is defined.</li>
         <li>
@@ -804,7 +804,7 @@
           <a data-cite="YAML#node-anchors">node anchor</a></var>,
           add a reference from the anchor name to the
           <a data-lt="YAML sequence">sequence</a>
-          in the <var>named nodes</var> map.</li>
+          in <var>named nodes dictionary</var>.</li>
         <li>For each <a>node</a> <var>n</var> in the
           <a data-lt="YAML sequence">sequence</a>,
           append the result of processing <var>n</var>
@@ -832,7 +832,7 @@
           <a data-cite="YAML#node-anchors">node anchor</a></var>,
           add a reference from the anchor name to
           the <a data-lt="YAML mapping">mapping</a>
-          in the <var>named nodes</var> map.</li>
+          in <var>named nodes dictionary</var>.</li>
         <li>For each <var>entry</var> in the <a data-lt="YAML mapping">mapping</a>
           composed of a key/value pair:
           <ol>
@@ -860,7 +860,7 @@
 
       <p>
         The conversion result is the value of the entry
-        in <var>named nodes</var> having the <a>node</a> entry.
+        in <var>named nodes dictionary</var> having the <a>node</a> entry.
         If none exist, the document is invalid,
         and processing MUST end in failure.
       </p>
--- a/spec/index.html
+++ b/spec/index.html
@@ -344,14 +344,17 @@
       <dfn data-cite="YAML#representation-graph" data-no-xref="">YAML representation graph</dfn>,
       <dfn data-cite="YAML#streams" data-no-xref="" data-lt="stream">YAML stream</dfn>,
       <dfn data-cite="YAML#directives" data-no-xref="" data-lt="directive">YAML directive</dfn>,
-      <dfn data-cite="YAML#block-mappings" data-no-xref="" data-lt="YAML block mapping">block mapping</dfn>,
-      <dfn data-cite="YAML#block-sequences" data-no-xref="" data-lt="YAML block sequence">block sequence</dfn>,
-      <dfn data-cite="YAML#flow-mappings" data-no-xref="" data-lt="YAML flow mapping">flow mapping</dfn>,
-      <dfn data-cite="YAML#flow-sequences" data-no-xref="" data-lt="YAML flow sequence">flow sequence</dfn>,
+      <dfn data-cite="YAML#sequence" data-no-xref="">YAML sequence</dfn>
+      (either
+      <dfn data-cite="YAML#block-sequences" data-no-xref="">block sequence</dfn> or
+      <dfn data-cite="YAML#flow-sequences" data-no-xref="">flow sequence</dfn>),
+      <dfn data-cite="YAML#mapping" data-no-xref="">YAML mapping</dfn>
+      (either
+      <dfn data-cite="YAML#block-mappings" data-no-xref="">block mapping</dfn> or
+      <dfn data-cite="YAML#flow-mappings" data-no-xref="">flow mapping</dfn>),
       <dfn data-cite="YAML#nodes" data-no-xref="" data-lt="YAML node">node</dfn>,
       <dfn data-cite="YAML#scalar" data-no-xref="" data-lt="YAML scalar">scalar</dfn>,
       <dfn data-cite="YAML#anchors-and-aliases" data-no-xref="">named anchor</dfn>,
-      <span class="ednote">(should this be <a data-cite="YAML#named-handles" data-no-xref="">named handle</a>?)</span>
       and <dfn data-cite="YAML#alias-nodes" data-no-xref="">alias node</dfn>,
       are imported from [[YAML]].</p>
 
@@ -506,7 +509,7 @@
       (see <a href="#int" class="sectionRef"></a>).
     </p>
 
-    <p>FIXME.</p>
+    <p class="ednote">FIXME.</p>
   </section>
 
   <section id="specifications" class="normative">
@@ -573,10 +576,11 @@
     </pre>
 
     <p>
-      While the representation graph (and eventually the in-memory representation
-      of the data structure, e.g., a Python dictionary or a Java hashmap) will still
-      contain references between nodes, the JSON-LD serialization will not, as shown
-      below:
+      While the representation graph
+      (and eventually the in-memory representation
+      of the data structure, e.g., a Python dictionary or a Java hashmap)
+      will still contain references between nodes,
+      the JSON-LD serialization will not, as shown below:
     </p>
 
     <pre class="example json"
@@ -648,6 +652,9 @@
       processed separately.
     </p>
 
+    <p class="ednote">
+      This is inconsistent with the processing description in <a href="#convert-stream" class="sectionRef"></a>.
+    </p>
   </section>
 
   <section id="conversion">
@@ -668,7 +675,8 @@
 
     <aside class="example" title="YAML Sequences">
       <p>For example, YAML has multiple ways to encode an <a>array</a>,
-        YAML <a>block sequences</a> and <a>flow sequences</a>.
+        YAML <a>block sequences</a> and
+        <a>flow sequences</a>.
         Both forms describe the same <a>array</a> of two <a>strings</a>.</p>
 
       <p>A YAML <a>block sequence</a>:</p>
@@ -721,6 +729,7 @@
         </li>
         <li>The conversion result is <var>stream content</var>.</li>
       </ol>
+
       <p>Any error reported in a recursive processing step results
         in the failure of this processing step.</p>
     </section>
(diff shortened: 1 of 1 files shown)


anatoly-scherbakov (dismissed):

spec/index.html
```diff
@@ -617,6 +633,132 @@ <h2>Streams</h2>
...
+
+    <section id="convert-document">
+      <h3>Converting a <a>YAML document</a></h3>
+
+      <p>From the <a data-cite="YAML#">YAML grammar</a>,
+        a YAML document MAY be proceeded by a
+        <a data-cite="YAML#rule-l-document-prefix">Document Prefix</a>
+        If a document contains a <a data-cite="YAML#rule-c-byte-order-mark">byte order mark</a>, it MUST
+        indicate the UTF-8 character encoding.
+        Otherwise a processing error has been found, and processing is aborted.</p>
```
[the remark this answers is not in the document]
gkellogg: In general, we're going to need to be more explicit on error cases, along the lines of the JSON-LD API. This may be considered an extension to that API, and integrating with that has its own challenges.

spec/index.html
```diff
@@ -617,6 +633,132 @@ <h2>Streams</h2>
...
+        <li><a>YAML block mapping</a>, or</li>
+        <li><a>YAML flow mapping</a>.</li>
+      </ul>
+      <p>Any error reported in a recursive processing step results
+        in the failure of this processing step.</p>
+    </section>
+
+    <section id="convert-block-seq">
+      <h3>Converting a <a>YAML block sequence</a></h3>
+    </section>
```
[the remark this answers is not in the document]
gkellogg: Yes, it is tedious. It would be better to rely on an external definition of such a transformation, since other than the use of the internal representation, this is fairly generic YAML processing.

@ioggstream has suggested raising an issue on the YAML media-type repo, although it's not clear that that is a proper place for it either.

The line I'm going on here is still relatively simple, but making sure that anchors, references, and tags are properly treated may push it too far.

spec/index.html
```diff
@@ -617,6 +633,132 @@ <h2>Streams</h2>
...
+    <section id="convert-block-map">
+      <h3>Converting a <a>YAML block mapping</a></h3>
+    </section>
+
+    <section id="convert-flow-map">
+      <h3>Converting a <a>YAML flow mapping</a></h3>
+    </section>
+
+    <section id="convert-scalar">
+      <h3>Converting a <a>YAML scalar</a></h3>
```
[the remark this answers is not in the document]
gkellogg: JSON only has numbers, while YAML has (at least) floats and ints. The basic profile would just use a generic number datatype. We might consider that the extended profile would maintain more fidelity. We can consider extending the internal representation with sub-types of number to handle these cases.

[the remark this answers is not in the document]
gkellogg: I'm going to do something different than what you propose here, though. The internal representation doesn't have quotes surrounding streams, only the YAML/JSON serialization forms.

spec/index.html
```diff
@@ -404,9 +420,9 @@ <h2>Basic Concepts</h2>
     <ul>
       <li>YAML supports different encodings, including UTF-8, UTF-16, and UTF-32.</li>
       <li>YAML supports more native data types than JSON.</li>
-      <li>The structure of a YAML document &mdash; that is, a named YAML representation graph &mdash;
+      <li>The structure of a <a>YAML document</a> &mdash; that is, a named <a>YAML representation graph</a> &mdash;
       is a rooted, directed graph that can have cycles.</li>
-      <li>YAML has the concept of a <a>YAML stream</a>, which is a sequence of documents.
+      <li>YAML has the concept of a <a>YAML stream</a>, which is a sequence of <a>YAML documents</a>.
```
TallTed, line 425:
```suggestion
      <li>YAML (1.2.2 and later) has the concept of a <a>YAML stream</a>, which is a sequence of <a>YAML documents</a>.
```

gkellogg:
This should be complete now. I'm sure it's inadequate in its treatment of tags. I think the named anchors and anchor nodes are handled consistently with the YAML spec, though.

This needs a careful review from those more familiar with the details of YAML.

spec/index.html
```diff
@@ -631,6 +652,247 @@ <h2>Streams</h2>
...
+    <p>YAML-LD processing is defined by converting YAML to the
+      <a>internal representation</a> and using [[[JSON-LD11-API]]]
+      to process on that representation,
+      after which the representation is converted back to YAML.
+      As information specific to a given <a>YAML document</a> structure is lost
+      in this transformation, much of the specifics of that
+      original representation are therefore lost in that conversion,
+      limiting the ability to fully round-trip a YAML-LD document back
+      to an equivalent representation.
+      Consequently, round-tripping is intended to preserve the semantic
```
TallTed, line 672:
```suggestion
      Consequently, round-tripping in this context is limited to preservation of the semantic
```

spec/index.html
```diff
@@ -631,6 +652,247 @@ <h2>Streams</h2>
...
+        <a data-cite="YAML#rule-l-directive">directives</a>
+        followed by a
+        <a data-cite="YAML#rule-l-bare-document">YAML bare document</a>,
+        which is composed of a single <a>node</a>.
+      </p>
+
+      <ol>
+        <li>Create an empty may <var>named nodes</var>,
+          used the associate an anchor name with the <a>node</a>
+          on which it is defined.</li>
```
TallTed, line 753: Guessing wildly at what was intended to be said here...
```suggestion
        <li>Create an empty <var>named node</var>,
          and associate an anchor name with the <a>node</a>
          on which it is defined.</li>
```

spec/index.html
```diff
@@ -631,6 +652,247 @@ <h2>Streams</h2>
...
+        <li>
+          If a document contains a
+          <a data-cite="YAML#rule-c-byte-order-mark">byte order mark</a>,
+          it MUST indicate the UTF-8 character encoding.
+          Otherwise, a processing error MUST be returned,
+          and processing is aborted.
```
TallTed, line 759:
```suggestion
          and processing aborted.
```

spec/index.html
```diff
@@ -631,6 +652,247 @@ <h2>Streams</h2>
...
+
+      <ol>
+        <li>Set <var>sequence content</var> to an empty <a>array</a>.</li>
+        <li>If the <a data-lt="YAML sequence">sequence</a> has a
+          <a data-cite="YAML#node-anchors">node anchor</a></var>
+          add a reference from the anchor name to the
+          <a data-lt="YAML sequence">sequence</a>
+          in the <var>named nodes</var> map.</li>
+        <li>For each <a>node</a> <var>n</var> in the
+          <a data-lt="YAML sequence">sequence</a>:
```
TallTed, line 809:
```suggestion
          <a data-lt="YAML sequence">sequence</a>,
```

spec/index.html
```diff
@@ -631,6 +652,247 @@ <h2>Streams</h2>
...
+      <h3>Converting a <a>YAML mapping</a></h3>
+
+      <p>Both <a>block mappings</a> and <a>flow mappings</a>
+        are directly aligned with
+        a <a>map</a> in the <a>internal representation</a>.</p>
+
+      <ol>
+        <li>Set <var>mapping content</var> to an empty <a>map</a>.</li>
+        <li>Otherwise, if the <a data-lt="YAML mapping">mapping</a> has a
+          <a data-cite="YAML#node-anchors">node anchor</a></var>
```
TallTed, line 832:
```suggestion
          <a data-cite="YAML#node-anchors">node anchor</a></var>,
```

spec/index.html
```diff
@@ -631,6 +652,247 @@ <h2>Streams</h2>
...
+          in the <var>named nodes</var> map.</li>
+        <li>For each <var>entry</var> in the <a data-lt="YAML mapping">mapping</a>
+          composed of a key/value pair:
+          <ol>
+            <li>Set <var>key</var> and <var>value</var>
+              to the result of processing <var>entry</var>
+              using the appropriate conversion step.</li>
+            <li>If <var>key</var> is not a <a>string</a>,
+              a processing error MUST be returned,
+              and processing is aborted.</li>
```
TallTed, line 844:
```suggestion
              and processing aborted.</li>
```

spec/index.html
gkellogg: I sure screwed up that sentence! 

At this point, we don't have a node to create such an association, only a `map` used for saving them in deeper steps.

It should have been the following:

```suggestion
        <li>Create an empty <a>map</a> <var>named nodes</var>,
          used the associate an anchor name with the <a>node</a>
          on which it is defined.</li>
```

spec/index.html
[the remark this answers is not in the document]
gkellogg: We will/should of course have comprehensive tests for these things, but we can't rely on tests to specify normative behavior. We need unambiguous description of the transformation process, at least into our _internal representation_.

That said, I'd much rather refer to a process than specify it (see https://github.com/ietf-wg-httpapi/mediatypes/issues/59).

spec/index.html
TallTed: @gkellogg — That helps with the first line, but the two that follow still need some changes to make the whole sentence coherent.

spec/index.html
```diff
@@ -631,6 +652,259 @@ <h2>Streams</h2>
...
+        <a data-cite="YAML#rule-l-directive">directives</a>
+        followed by a
+        <a data-cite="YAML#rule-l-bare-document">YAML bare document</a>,
+        which is composed of a single <a>node</a>.
+      </p>
+
+      <ol>
+        <li>Create an empty <a>map</a> <var>named nodes</var>,
+          used the associate an anchor name with the <a>node</a>
+          on which it is defined.</li>
```
gkellogg, line 753: @TallTed How about the following:

```suggestion
        <li>Create an empty <a>map</a> <var>named nodes</var>,
          which will be used to associate an <a>alias node</a> with the <a>node</a>
         having the corresponding <a>named anchor</a>.</li>
```

anatoly-scherbakov (approved):

ioggstream:
@gkellogg did you get some feedback from YAML folks on this topic?

gkellogg:
> @gkellogg did you get some feedback from YAML folks on this topic?

No, not yet.

One necessary improvement to my text is to provide some handling of looping alias nodes along the lines of "when resolving aliases implementations MUST check for recursive entry of a referenced node and fail."

As I said, I think a process such as this should be defined elsewhere and referenced, but we'll keep it here until something better comes along. Improvements and sanity check, of course, are always welcome.

spec/index.html
TallTed: I think that almost works, to the degree I understand it.  I'd make one more change — in the first line, the plural `<var>named nodes</var>` should be singular.

spec/index.html
```diff
@@ -868,6 +868,18 @@ <h3>Converting an <a>Alias Node</a></h3>
...
       <h3>Converting a <a>YAML scalar</a></h3>
+      
+      <ul>
+        <li>A `null` scalar is kept as-is;</li>
+        <li>An `integer` scalar is kept as-is;</li>
+        <li>
+        A string scalar is surrounded by double quotes, unless
+            <ul>
+              <li>it has been surrounded by double quotes already, in which case they are kept as-is,
+              <li>it has been surrounded by singular quotes, in which case they are replaced by double quotes, and any double quotes in the scalar itself are escaped with the backslash `\` character.</li>
```
TallTed, line 879:
```suggestion
        A string scalar is surrounded by double-quotes (`&#34;`),
        except in the following cases:
            <ul>
              <li>it has already been surrounded by double-quotes,
                  in which case they are kept as-is</li>
              <li>it has already been surrounded by single-quotes (`&#39;`),
                  in which case the surrounding single-quotes are replaced by
                  double-quotes, and any double-quote in the scalar itself is
                  escaped with the backslash character (`&#92;`)</li>
```

spec/index.html
gkellogg: > I think that almost works, to the degree I understand it. I'd make one more change — in the first line, the plural `<var>named nodes</var>` should be singular.

I don't quite see that, as the _map_ relates alias nodes with nodes in general, it is not a single _named node_. Maybe something like `named nodes map` would be clearer?

spec/index.html
gkellogg: Actually, this shouldn't be in here, and was likely merged in error. The YAML model describes different syntactic forms of strings, but this describing relies on the underlying internal representation of a string, which is not syntactic.

The entire bulleted section should be removed.

gkellogg: Removed in fa87c16a123503dcc7b677f2bf28c8b2d9c52cc0.

spec/index.html
gkellogg: Changed `named nodes dictionary` in e4fa1f10dfff5af1ae57e17e356a6bf36b5e06b3.

gkellogg: Sorry, more in bd3dffce19204c66b0a79037fcafd380e0480352.

spec/index.html
```diff
@@ -748,9 +748,10 @@ <h3>Converting a <a>YAML document</a></h3>
       </p>
 
       <ol>
-        <li>Create an empty <a>map</a> <var>named nodes</var>,
-          used the associate an anchor name with the <a>node</a>
-          on which it is defined.</li>
+        <li>Create an empty <a>map</a> <var>named nodes dictionary</var>,
```
TallTed, line 751: Removing the HTML tags —
```
Create an empty map named nodes dictionary,
```
— just doesn't read right to me. I'm sure it's close, but I'm not sure what to change.

spec/index.html
```diff
@@ -804,7 +805,7 @@ <h3>Converting a <a>YAML sequence</a></h3>
           <a data-cite="YAML#node-anchors">node anchor</a></var>,
           add a reference from the anchor name to the
           <a data-lt="YAML sequence">sequence</a>
-          in the <var>named nodes</var> map.</li>
+          in <var>named nodes dictionary</var>.</li>
```
TallTed, line 808:
```suggestion
          in the <var>named nodes dictionary</var>.</li>
```

spec/index.html
```diff
@@ -832,7 +833,7 @@ <h3>Converting a <a>YAML mapping</a></h3>
           <a data-cite="YAML#node-anchors">node anchor</a></var>,
           add a reference from the anchor name to
           the <a data-lt="YAML mapping">mapping</a>
-          in the <var>named nodes</var> map.</li>
+          in <var>named nodes dictionary</var>.</li>
```
TallTed, line 836:
```suggestion
          in the <var>named nodes dictionary</var>.</li>
```

spec/index.html
```diff
@@ -860,26 +861,20 @@ <h3>Converting an <a>Alias Node</a></h3>
 
       <p>
         The conversion result is the value of the entry
-        in <var>named nodes</var> having the <a>node</a> entry.
+        in <var>named nodes dictionary</var> having the <a>node</a> entry.
```
TallTed, line 864:
```suggestion
        in the <var>named nodes dictionary</var> having the <a>node</a> entry.
```

spec/index.html
gkellogg: It is consistent with how such variables are described in JSON-LD 1.1 API. For example, from the [Compaction Algorithm](https://www.w3.org/TR/json-ld11-api/#algorithm-6):

> create a new array <var>compacted types</var> initialized by  ...

If it was <var>x</var> instead of <var>named nodes dictionary</var>, would it sound any different?

> Create an empty `map` <var>x</var> used the associate an anchor name with the `node`.

Would changing it to "empty `map` named <var>named nodes dictionary</var>" be better? To me _named nodes dictionary_ is the variable, it's not really the name of the variable.

gkellogg:
Note the comment from b95391285a5eee8f94d0a495921bb5ff347e48a3:

> This may be largely a matter of opinion, for which there is no _right_ answer, as writing algorithmic text is not quite the same as normal English.

Still, we strive for perfection in all things...

spec/index.html
```diff
@@ -631,6 +652,254 @@ <h2>Streams</h2>
...
+        followed by a
+        <a data-cite="YAML#rule-l-bare-document">YAML bare document</a>,
+        which is composed of a single <a>node</a>.
+      </p>
+
+      <ol>
+        <li>Create an empty <a>map</a> <var>named nodes dictionary</var>,
+          which will be used to the associate an <a>alias node</a>
+          with the <a>node</a>
+          having the corresponding <a>named anchor</a>.</li>
```
TallTed, line 754: I think I'm getting closer to understanding what's meant. This may do the job, if it works for you --
```suggestion
        <li>Create an empty <a>map</a> for the <var>named nodes dictionary</var>
          which will be used to associate each <a>alias node</a>
          with the <a>node</a>
          having the corresponding <a>named anchor</a>.</li>
```

If that doesn't, this may (but won't be a github suggestion you can accept with a click), with or without `dictionary` before the first `</var>` --
```suggestion
        <li>Create an empty <var>named nodes</var> <a>map</a>
          which will be used to associate each <a>alias node</a>
          with the <a>node</a>
          having the corresponding <a>named anchor</a>.</li>
```

gkellogg: Thanks @TallTed, yes I think the second is better.

gkellogg:
This issue was discussed on the [Aug 03](https://json-ld.org/minutes/2022-08-03/) meeting. and we agreed to merge.

PR #69, opened 2022-08-02
Fix #68. Cleanup notes.

### LINKED ISSUE #68: YAML streams pre-dates 1.2
- [x] See https://yaml.org/spec/1.1/#stream/information%20model I'd avoid too many informative notes: users MUST read YAML spec.
- [x] Since future YAML versions  could introduce dependencies between different documents in a stream, it's best to reference the Interoperability Considerations in the media type.

ioggstream:
## This PR

- fixes some issues in the document notes.

See #68

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/69.html" title="Last updated on Aug 5, 2022, 4:58 PM UTC (d334516)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/69/d1b039d...d334516.html" title="Last updated on Aug 5, 2022, 4:58 PM UTC (d334516)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -1025,7 +1025,7 @@
       <p>Fragment identifiers used with <a href="#application-ld-yaml">application/ld+yaml</a>
        are treated as in RDF syntaxes, as per
        <a data-cite="RDF11-CONCEPTS#section-fragID">RDF 1.1 Concepts and Abstract Syntax</a>
-       [[RDF11-CONCEPTS]].
+       [[RDF11-CONCEPTS]]
        and do not follow the process defined for application/yaml.
       </p>
 
--- a/spec/index.html
+++ b/spec/index.html
@@ -23,9 +23,9 @@
     localBiblio: {
       "I-D.ietf-httpapi-yaml-mediatypes": {
         title: "YAML Media Type",
-        href: "https://datatracker.ietf.org/doc/draft-ietf-httpapi-yaml-mediatypes/",
+        href: "https://www.ietf.org/archive/id/draft-ietf-httpapi-yaml-mediatypes-03.html",
         publisher: "IETF",
-        date: "2022-06-23",
+        date: "2022-08-05",
         status: "WG Document",
         authors: [
           "Roberto Polli",
@@ -293,11 +293,10 @@
       of <a>YAML-LD documents</a>.
     </p>
 
-    <p class="note" title="Documents in the stream are independent of each other">
-        Each document in a stream has its own context, [[YAML]] directives, <a>node anchors</a>, and so on.
+    <p class="note" title="Interoperability considerations on YAML streams">
+        For interoperability considerations on YAML streams,
+        see <a data-cite="I-D.ietf-httpapi-yaml-mediatypes#section-3.2">the relevant section in YAML Media Type</a>.
     </p>
-
-    <p class="note"><a>YAML streams</a> were introduced in YAML 1.2.2, and cannot be used in earlier versions.</p>
     
     <p>A <dfn>YAML-LD document</dfn> is any <a>YAML document</a> from
       which a conversion to [[JSON]] produces
@@ -344,6 +343,7 @@
       <dfn data-cite="YAML#representation-graph" data-no-xref="">YAML representation graph</dfn>,
       <dfn data-cite="YAML#streams" data-no-xref="" data-lt="stream">YAML stream</dfn>,
       <dfn data-cite="YAML#directives" data-no-xref="" data-lt="directive">YAML directive</dfn>,
+      <dfn data-cite="YAML#documents" data-no-xref="" data-lt="documents">YAML document</dfn>,
       <dfn data-cite="YAML#sequence" data-no-xref="">YAML sequence</dfn>
       (either
       <dfn data-cite="YAML#block-sequences" data-no-xref="">block sequence</dfn> or
@@ -357,10 +357,6 @@
       <dfn data-cite="YAML#node-anchors" data-lt="anchor name|anchored nodes" data-no-xref="">node anchor</dfn>,
       and <dfn data-cite="YAML#alias-nodes" data-no-xref="">alias node</dfn>,
       are imported from [[YAML]].</p>
-
-    <p>The term
-      <dfn>YAML document</dfn> represents a serialization of a resource
-      conforming to the [[YAML]] grammar.</p>
 
     <p>The term
       <dfn data-cite="RFC9110#content.negotiation" data-no-xref="">content negotiation</dfn>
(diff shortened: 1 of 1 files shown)


gkellogg:
Much more useful to reference the ID, but it doesn’t seem to have anything to say about YAML streams. If that forthcoming?

gkellogg:
Fixes #68,

spec/index.html
```diff
@@ -294,10 +294,9 @@ <h2>Terminology</h2>
     </p>
 
     <p class="note" title="Documents in the stream are independent of each other">
-        Each document in a stream has its own context, [[YAML]] directives, named anchors, and so on.
+        For interoperability considerations on YAML streams,
+        see the YAML Streams section in [[I-D.ietf-httpapi-yaml-mediatypes]].
```
gkellogg, line 298:
```suggestion
        For interoperability considerations on <a>YAML streams<a>,
         see the YAML Streams section in [[I-D.ietf-httpapi-yaml-mediatypes]]
         <div class="ednote">Update to a data-cite reference when the ID has a YAML Streams section</div>.
```

spec/index.html
```diff
@@ -621,11 +619,11 @@ <h2>Anchors and aliases</h2>
...
       <h2>Streams</h2>
 
-      <p class="note">YAML streams were introduced in YAML 1.2.2, and cannot be used in earlier versions.</p>
 
       <p>
-        A single YAML-LD file might contain multiple <a>YAML-LD documents</a>
-        using the mechanism of <a>YAML-LD streams</a>, as shown in the example below.
+        Since a single YAML-LD file is a <a>YAML-LD stream</a>,
+        it might contain multiple <a>YAML-LD documents</a>,
+        as shown in the example below.
```
gkellogg, line 626: The wording can be improved here. I suspect @TallTed will comment.

spec/index.html
```diff
@@ -648,7 +646,8 @@ <h2>Streams</h2>
     </section>
 
     <p>
-      Each of the individual <a>YAML documents</a> in the stream has to be converted into a <a>JSON-LD document</a> and
+      Each of the individual <a>YAML documents</a> in the stream
+      has to be converted into a separated <a>JSON-LD document</a> and
```
gkellogg, line 650: Not that I necessarily agree with this requirement, but I think this sounds better.

```suggestion
      MUST be converted into a separate <a>JSON-LD document</a> and
```

gkellogg (changes requested):
Some changes for style, but not really content. I think we need to figure out how we want to handle multiple documents in a stream, but this is a good default position.

spec/index.html
ioggstream: This is the link https://github.com/ietf-wg-httpapi/mediatypes/blob/main/draft-ietf-httpapi-yaml-mediatypes.md#yaml-streams-int-yaml-streams but I don't know how to reference it via [[xxx]]

spec/index.html
ioggstream: I think it's not a requirement, but the only thing you can do. While you can convert the YAML-LD stream to a JSON sequence, every document can only be converted to a separate JSON document.

We could define a way to "merge" all the YAML-LD documents in a stream into a single JSON-LD document, but this is quite complex to do reliably.

spec/index.html
gkellogg: That link seems to be to an unpublished draft. The normative reference in our spec is https://datatracker.ietf.org/doc/draft-ietf-httpapi-yaml-mediatypes/. Once the ID is updated with the section on streams, we would change this to `<a data-cite=-D.ietf-httpapi-yaml-mediatypes#section-3.4">YAML Streams</a>`.

Actually, the cited document should probably be the HTML version at https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-yaml-mediatypes-02, rather than the otherwise opaque version we use currently, which exposes the section identifiers, such as `#section-3.1`. So, I'd recommend updating the localBibliography entry to use that URL instead of the current one.

Any time-frame for  a 3rd draft of the ID?

gkellogg: This is the localBiblio entry containing the href which should be updated:

https://github.com/json-ld/yaml-ld/blob/d1b039d37a1333276d341f8d341d7b2494741f0b/spec/index.html#L24-L35

spec/index.html
gkellogg: I think we discussed that this is similar to how the [JSON-LD API HTML content algorithms](https://www.w3.org/TR/json-ld11-api/#html-content-algorithms) deal with multiple script elements. Specifically, step 6.2:

> Otherwise, the [extractAllScripts](https://www.w3.org/TR/json-ld11-api/#dom-loaddocumentoptions-extractallscripts) option is true. Set document to a new empty [array](https://infra.spec.whatwg.org/#list). For each [JSON-LD script element](https://www.w3.org/TR/json-ld11-api/#dfn-json-ld-script-element) in input:

> 1. Set source to its [textContent](https://dom.spec.whatwg.org/#dom-node-textcontent).
> 2. Set script content to the result of the [Extract Script Content algorithm](https://www.w3.org/TR/json-ld11-api/#extract-script-content), using source, rejecting promise with a [JsonLdError](https://www.w3.org/TR/json-ld11-api/#dom-jsonlderror) whose code set from the result, if an error is detected and processing is terminated.
> 3. If script content is an [array](https://infra.spec.whatwg.org/#list), merge it to the end of document.
> 4. Otherwise, append script content to document.

spec/index.html
TallTed, line 626:
```suggestion
        Every YAML-LD file is a <a>YAML-LD stream</a>
        and might contain multiple <a>YAML-LD documents</a>,
        as shown in the example below.
```

spec/index.html
ioggstream: done

spec/index.html
ioggstream: I'll file another issue and leave the current text as is in this PR.

spec/index.html
```diff
@@ -1031,17 +1026,10 @@ <h3>Fragment identifiers</h3>
        are treated as in RDF syntaxes, as per
        <a data-cite="RDF11-CONCEPTS#section-fragID">RDF 1.1 Concepts and Abstract Syntax</a>
        [[RDF11-CONCEPTS]].
+       and do not follow the process defined for application/yaml.
       </p>
 
-      <p>
-        The YAML media type supports both <a>alias nodes</a> and JSON Pointers [[RFC6905]]
```
ioggstream, line 1037: @gkellogg we don't use the native fragment identifier, so we don't need to explain it.

spec/index.html
```diff
@@ -1031,17 +1026,10 @@ <h3>Fragment identifiers</h3>
        are treated as in RDF syntaxes, as per
        <a data-cite="RDF11-CONCEPTS#section-fragID">RDF 1.1 Concepts and Abstract Syntax</a>
        [[RDF11-CONCEPTS]].
```
gkellogg, line 1028:
```suggestion
       [[RDF11-CONCEPTS]].
```

gkellogg (approved):

Issue #75, opened 2022-08-17
Work Plan


gkellogg:
The general work plan adopted on the [August 17th](https://json-ld.org/minutes/2022-08-17/) meeting:

* [x] Define limitations on use of YAML Alias Nodes.
  * Use of aliases limited to the Extended profile.
* [x] Define more specifically the YAML-LD-JSON Profile. (PR yaml-ld#70)
* [x] Define YAML-LD-Extended Profile.
  * [x] Transparent use of %TAG for XSD and I18N namespaces (pending library support) (issue yaml-ld#17)
  * [x] Full use of YAML Alias nodes.  Not round-tripped.
* [x] Define extension points to JSON-LD API.
* [x] Test Suite (see #20)

This issue serves as a meta-issue to manage completion of different milestones.

Issue #55, opened 2022-07-07
Choose prefix character to replace @


anatoly-scherbakov:
As an architect of a YAML-LD aware system and a developer, I am a bit uncomfortable about having my users write JSON-LD @-keywords and have to quote them because that would mean additional effort. I would rather prefer some other prefix character from ASCII character set.

## Choice of prefix characters

| @id | Notes | Resolution
| --- | --- | --- |
| `@id` | Is [reserved](https://yaml.org/spec/1.2.2/#rule-c-directive). | x |
| `id | Is [reserved](https://yaml.org/spec/1.2.2/#rule-c-directive). | x |
| ~id | Works |  |
| !id | Used for [tags](https://yaml.org/spec/1.2.2/#tag-handles) | x |
| #id | Used for [comments](https://yaml.org/spec/1.2.2/#rule-c-nb-comment-text) | x |
| $id | Works | x |
| %id | Used for [directives](https://yaml.org/spec/1.2.2/#rule-c-directive) | x |
| ^id | Works |  |
| &id | Used for [anchors](https://yaml.org/spec/1.2.2/#anchors-and-aliases) | x |
| *id | Used for [aliases](https://yaml.org/spec/1.2.2/#anchors-and-aliases) | x |
| -id | Works, but is confusable with lists: `- id` | x |
| +id | Works |  |
| _id | Works but is confusable with blank nodes (#57) | x |

Also, `_` in Python means private or internal things so I am subconsciously against that :)

I suggest using `$` for the Convenience Context. This character also is good for JS: `something.$type` is workable in that language.

The users however may do whatever they want. For instance, users might want something like this:

```yaml
=: 🏠:cat
∊: 🐱
🏷️:
  🇺🇸: Ray
```

I won't stop them. Our `$` is really just a way to trigger everyone's imagination.

VladimirAlexiev:
A good analysis what chars are suitable for #51

gkellogg:
Discussed in the TPAC F2F and [resolved to close as won't fix](https://json-ld.org/minutes/2022-09-12/#resolution-2).

Issue #13, opened 2022-05-30
Define anchor usage in yaml-ld


ioggstream:
As an json-ld editor      … WHO
I want to use yaml anchors … WHAT
So that I  can easily reuse content  … WHY

## Note

The specification should define:

- when it is legitimate to use anchors
- which are the expectation on anchor usage (e.g. do they represent a specific JSON-LD node or they can just be used to represent content?)
- are there any constraint on anchor usage? (e.g. the representation graph MAY / MUST NOT be a cyclic graph...)

## example 1

```
---
- "@id": &homer http://example.org/#homer
  http://example.com/vocab#name:
  - "@value": Homer
- "@id": http://example.org/#bart
  http://example.com/vocab#name:
  - "@value": Bart
  http://example.com/vocab#parent:
  - "@id": *homer
- "@id": http://example.org/#lisa
  http://example.com/vocab#name:
  - "@value": Lisa
  http://example.com/vocab#parent:
  - "@id": *homer
```
## example 2 

Using anchor and alias nodes https://gist.github.com/ioggstream/31f3226fa9976b3baf0800f44bc19c98

pchampin:
One point where I believe YAML anchors can help are the description complex of contexts. E.g.
```
{
  "@context": {
    "xsd": "http://www.w3.org/2001/XMLSchema#",
    "@vocab": "http://example.com/ns/Company/",
    "founder": { "@context": {
        "@vocab": "http://example.com/ns/Person/",
        "birthDate": { "@type": "xsd:date" }
    }},
    "employee": { "@context": {
        "@vocab": "http://example.com/ns/Person/",
        "birthDate": { "@type": "xsd:date" }
    }}
  }
}
```
Notice that the scoped contexts of `founder` and `employee` are exactly the same (a "person" context). With Yaml anchors, this redundancy could be elimiinated.

NB: there are other means to get rid of this redundancy in pure JSON-LD:
* hosing the "person" context at a different URL and use that URL instead
* define a type-scoped context for a type `Person`, and expect values of `founder` and `employee` to be explicitly typed

but they have their drawbacks that are not always acceptable.

VladimirAlexiev:
>  how Anchors and Aliases could mesh with JSON-LD Frames

Frames specify which nodes to expand, and which nodes to merely refer to by URI.
So in some sense they tackle the "graph vs tree" problem.

Anchors and Aliases tackle the same problem; intuitively I feel in a more general way.

So: what can be the connection between them?

[...]

Issue #11, opened 2022-05-29
Replace $-keywords with @-keywords


anatoly-scherbakov:
As an `author of YAML-LD files`       … WHO
I want an `ability to type keywords without quotes`  … WHAT
So that `my authoring experience is better`   … WHY

# Motivation

I believe the primary purpose of having a Linked Data format based on YAML is to simplify manual authoring of the linked data documents. This means that, in an information system, we could ask domain experts to write YAML documents to describe their knowledge.

YAML is much easier to write manually than JSON because it does not require as much syntactic noise. Normally, keys can be written without quoting at all:

```yaml
date: 2022-05-29
title: This is my latest blog article
```

However, sooner or later the document author will have to define `@type`, `@context`, `@language`, or any other JSON-LD keyword; and then they have to remember that `@` is a reserved character and that in such cases quoting is mandatory.

- This is an edge case which requires cognitive effort,
- And it requires extra time to type.

The potential author of YAML-LD documents is not necessarily a programmer; they might be a history student, an anthropologist, a biologist, a physicist.

Let's not make their life harder than it has to be.

# Potential risks

- The `@` character might suddenly acquire a new meaning under YAML specification. — This is not a risk because, under my proposal, we'll not have `@` in our documents anyway, we'll have `$`.
- The `$` character might suddenly acquire a new meaning under YAML specification. — I consider that unlikely: it is now not, and such a change would bear great risks for broken backwards compatibility, especially for those who combine YAML with JSON Schema.
- YAML-LD keywords will intersect with JSON Schema keywords. — I believe the sets of keywords between JSON-LD and JSON Schema do not intersect. For instance, JSON-LD does not have `@schema` or `@ref`, and JSON Schema does not have `$context` or `$type`. I am saying this based on a brief web research, please do correct me if I am wrong.

# Possible implementations

- Not replace anything, as per #9: possible, but I have been writing YAML-LD documents for some time now and my subjective feeling is that the replacement really had improved by writing experience. Not a good argument though as this is very much IMHO.
- `!yaml-ld` tag proposed at #6: interesting, but for a non-programmer it will be just a nuisance, a piece of syntactic noise. I would propose to prioritize it to make YAML-LD as concise as possible without losing its writeability and readability.

# Proposal

Let us replace `$` → `@` and vice versa only for the particular keywords. For instance,

```yaml
$schema: boo
$context: foo
```

will be converted into

```
{
  "$schema": "boo",
  "@context": "foo"
}
```

[...]

PR #77, opened 2022-08-20
Definition of yaml-ld namespace documents

gkellogg:
… and updates to the spec to…refer and use it.

For #13, #55, and #75.

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/77.html" title="Last updated on Aug 23, 2022, 3:50 PM UTC (c95ef08)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/77/6821734...c95ef08.html" title="Last updated on Aug 23, 2022, 3:50 PM UTC (c95ef08)">Diff</a>


gkellogg:
This adds some framing keywords to the previous version discussed in the Best Practice section, and uses a versioned URI based on the yaml-ld namespace. It also defines the namespace documents (html and jsonld).

Once committed, will need @pchampin's aid in putting the files in place along with appropriate .htaccess for content negotiation.

README.md
```diff
@@ -8,6 +8,8 @@ as developed by the [JSON for Linking Data Community Group](https://www.w3.org/c
 
 The Use Cases and Requirements document can also be read  [read directly](https://json-ld.github.io/yaml-ld/UCR/).
 
+The YAML-LD Namespace contains a [convenience context](https://www.w3.org/ns/yaml-ld/v1) which can also be [red directly](https://json-ld.github.io/yaml-ld/ns/v1.html).
```
pchampin, line 11:
```suggestion
The YAML-LD Namespace contains a [convenience context](https://www.w3.org/ns/yaml-ld/v1) which can also be [read directly](https://json-ld.github.io/yaml-ld/ns/v1.html).
```

pchampin (approved):
LGTM, modulo a typo

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
...
+
+    <section class="informative" id="convenience-context">
+      <h3>Convenience Context</h3>
+      <p>As JSON-LD keywords cannot be represented in YAML without being quoted,
+        publishers may find it useful to declare term aliases for these
+        keywords.</p>
+
+      <p>
+        For convenience, a <dfn>convenience context</dfn> is available
+        at `https://www.w3.org/ns/yaml-ld/v1`, which pre-defines these aliases:
```
TallTed, line 527:
```suggestion
        To make this easier, a <dfn>convenience context</dfn> which pre-defines
        these aliases is available at `https://www.w3.org/ns/yaml-ld/v1`, and
        partially shown below:
```

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
...
+            "$value": "@value",
+            "$version": "@version",
+            "$vocab": "@vocab"
+          }
+        }
+        -->
+      </pre>
+
+      <p>The <a>convenience context</a> contains an alias to every JSON-LD keyword which the JSON-LD 1.1
+        specification permits aliasing &mdash; which means all the keywords except <code>@context</code>. The reserved `@` character is
```
TallTed, line 570:
```suggestion
        specification permits to be aliased &mdash; which means all the keywords except <code>@context</code>. The reserved `@` character is
```

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
...
+        replaced by `$`, which is not reserved and therefore does not require quoting. Consider
+        <a href="#example-example-yaml-ld-document-without-convenience-context"></a>
+         reformatted using the <a>convenience context</a>:</p>
```
TallTed, line 573: The latter sentence isn't quite a sentence. I think there should probably be something in the `<a />` tag?

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
...
+
+    <p>This allows YAML-LD to be created with minimal use of <a>map</a> quoted keys (Note that the `@context` keyword cannot be aliased):</p>
```
TallTed, line 575: The "note" is redundant to a couple paragraphs above, and the "minimal use" sentence doesn't quite flow from the paragraph immediately preceding...

spec/index.html
```diff
@@ -661,6 +739,18 @@ <h2>Streams</h2>
     <p class="ednote">
       This is inconsistent with the processing description in <a href="#convert-stream" class="sectionRef"></a>.
     </p>
+    </section>
+
+  <section id="pre-loaded-contexts">
+    <h3>Pre-loaded Contexts</h3>
+    <p>In order to reduce the demand for loading static documents,
```
TallTed, line 746:
```suggestion
    <p>To reduce the demand for loading static documents,
```

spec/index.html
```diff
@@ -661,6 +739,18 @@ <h2>Streams</h2>
...
+      implementations SHOULD maintain a locally cached version of the following
+      documents the be satisfied by the default <a>documentLoader</a>.</p>
```
TallTed, line 748: maybe?
```suggestion
      documents that will be satisfied by the default <a>documentLoader</a>.</p>
```

TallTed (changes requested):

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
     </p>
 
     <p class="ednote">FIXME.</p>
+
+    <section class="informative" id="convenience-context">
+      <h3>Convenience Context</h3>
```
ioggstream, line 520: :-1:

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
...
+            "$version": "@version",
+            "$vocab": "@vocab"
+          }
+        }
+        -->
+      </pre>
+
+      <p>The <a>convenience context</a> contains an alias to every JSON-LD keyword which the JSON-LD 1.1
+        specification permits aliasing &mdash; which means all the keywords except <code>@context</code>. The reserved `@` character is
+        replaced by `$`, which is not reserved and therefore does not require quoting. Consider
```
ioggstream, line 571: Another sign that  this convenience context does not belong here: `@context` with `@`, other keywords with `$` and then more words and caveats.

spec/index.html
```diff
@@ -515,20 +515,99 @@ <h2>Basic Concepts</h2>
...
+         data-transform="updateExample"
+         data-content-type="application/ld+yaml"
+         title="Basic YAML-LD document with convenience context">
+      <!--
+      %YAML 1.2
+      ---
+      '@context':
+      - http://example.org/context.jsonld
+      - https://www.w3.org/ns/yaml-ld/v1
+      $graph:
```
ioggstream, line 587: Does this mean that if I'm offline I cannot even use a JSON-LD processor to understand that `$graph` is a `@graph` ?

spec/index.html
```diff
@@ -661,6 +739,18 @@ <h2>Streams</h2>
     <p class="ednote">
       This is inconsistent with the processing description in <a href="#convert-stream" class="sectionRef"></a>.
     </p>
+    </section>
+
+  <section id="pre-loaded-contexts">
```
ioggstream, line 744: Are those considerations specific to YAML? If not, I'd put there in another document. If yes, I think we are making things more complex than they should.

README.md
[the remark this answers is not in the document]
gkellogg: Fixed in 314b84df664448f948df3224ca27336304b90388.

ns/v1.haml
```diff
@@ -0,0 +1,87 @@
+<html lang='en'>
```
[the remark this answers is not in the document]
gkellogg: Fixed in 314b84df664448f948df3224ca27336304b90388.

README.md
gkellogg: Fixed in 314b84df664448f948df3224ca27336304b90388.

spec/index.html
gkellogg: When the content of the anchor with an internal reference is blank, ReSpec fills in the title of that element, the rendered text reads as follows:

> Consider [Example 12](http://localhost:8000/spec/#example-example-yaml-ld-document-without-convenience-context) reformatted using the [convenience context](http://localhost:8000/spec/#dfn-convenience-context):

But, yes, text moved around and this needs to be updated.

spec/index.html
gkellogg: There is text elsewhere (4.5 Pre-loaded Contexts) that this should be built into the `documentLoader`, so it is always available. It's a SHOULD, but perhaps should be a MUST.

Note that, this is the _default_ `documentLoader`, which can still be updated when invoking the API and it would be up to those implementations to continue to pre-load the context.

spec/index.html
gkellogg: There is general advice in the JSON-LD Best Practices document about caching contexts, and this falls under that guidance. This spec more strongly suggests that this particular context be implemented in the document loader. We can't update the core JSON-LD API spec for this right now.

If placed in a separate document (also non-normative, at this point), we'd need to add language here that implementations MUST adhere to those requirements, and we can't go back in time to require that JSON-LD processors adhere to the same requirements.

spec/index.html
TallTed: @ioggstream — It's unclear what you're saying you don't like.  Please elaborate.

gkellogg:
> I do not advise publishing a context which is specific to YAML.

The context is in the YAML-LD namespace, but of course, it could be used with any JSON-LD processor. Is there a more appropriate namespace to publish this? We did discuss `https://json-ld.org/contexts/`, but that does not have the same guarantees of permanent availability as the W3C namespace. Note that Verifiable Credientials (which is a Working Group, and the rules are different) uses a dated URL `https://www.w3.org/2018/credentials/v1`.

> Moreover, this overloads the dollar sign, too...

??? It does not overload the dollar sign, just defines terms that begin with the dollar sign. This seemed to be the group consensus, other than some use of non-ascii characters instead.

If there's still controversy on using the `$keywords`, we'll hold of merging and discuss on the next call.

> See the rest of the document for further comments.

spec/index.html
gkellogg: Since using a context to provide keyword alternates seemed to be a group consensus (I could be wrong, of course), this seems to be the natural consequence of that decision.

In #11, @anatoly-scherbakov laid out a plan for broad replacement of `@-keywords` with `$-keywords`, which would require deeper changes to document loading or core algorithms, and is still fraught with unexpected consequences. I outlined this solution in https://github.com/json-ld/yaml-ld/issues/11#issuecomment-1141352994.

@juusoautiosalo suggested in https://github.com/json-ld/yaml-ld/issues/11#issuecomment-1177200927 to chose a character other than `$`, and if there is consensus over some specific character, we can of course update this PR to use that, instead.

Of course, the ongoing need to quote `"@context"` is not solved by this solution, but seems to be a reasonable tradeoff for using an already standard mechanism for aliasing keywords.

spec/index.html
gkellogg: As I argued elsewhere, a heavier-weight mechanism for alternative keywords can usher in unintended consequences; the keyword aliasing mechanism provided already by contexts is the right solution, IMHO.

If the group can't reach consensus, we can simply document the issue and leave it for a future working group to address and provide no alternative to `@-keywords`.

TallTed:
> [@gkellogg] [We did discuss `https://json-ld.org/contexts/`, but that does not have the same guarantees of permanent availability as the W3C namespace](https://github.com/json-ld/yaml-ld/pull/77#issuecomment-1222667635)

`https://json-ld.org/contexts/` could potentially be used as the "actual" home of the documents, with `w3id.org` holding the permalinks through whatever aliases make sense, as the latter can be re-redirected if the "actual" home needs to change for whatever reason ...

spec/index.html
gkellogg: Please consider updated wording of the paragraph.

spec/index.html
```diff
@@ -567,13 +567,18 @@ <h3>Convenience Context</h3>
         -->
       </pre>
 
-      <p>The <a>convenience context</a> contains an alias to every JSON-LD keyword which the JSON-LD 1.1
-        specification permits to be aliased &mdash; which means all the keywords except <code>@context</code>. The reserved `@` character is
-        replaced by `$`, which is not reserved and therefore does not require quoting. Consider
-        <a href="#example-example-yaml-ld-document-without-convenience-context"></a>
-         reformatted using the <a>convenience context</a>:</p>
+      <p>The <a>convenience context</a> contains an alias for every JSON-LD keyword
+        the JSON-LD 1.1 specifications permit to be aliased &mdash;
```
TallTed, line 571:
```suggestion
        permitted to be aliased by the JSON-LD 1.1 specifications;
```

spec/index.html
gkellogg: In the JSON- LD API, it is described here: https://www.w3.org/TR/json-ld11-api/#remote-document-and-context-retrieval. The option to set it is `documentLoader`. It is the general way that documents are retrieved and interpreted.

spec/index.html
ioggstream: > We can't update the core JSON-LD API spec for this right now.

Can't we schedule an update of that document (or just update the draft version, if any?)


> This spec more strongly suggests that this particular context be implemented in the document loader

So we need this only to support the convenience context?

ioggstream:
> > I do not advise publishing a context which is specific to YAML. 
> The context is in the YAML-LD namespace,

I just meant that I expected YAML-LD to be interchangeable with JSON-LD without the need of bearing specific knowledge around.

> If there's still controversy on using the `$keywords`, we'll hold of merging and discuss on the next call.

I am really sorry for appearing so assertive on this point, but I did not imagine the impacts of all this stuff were so wide as to affect implementations too.

Thank you all for your patience,
R.

TallTed:
Re: https://github.com/json-ld/yaml-ld/pull/77#issuecomment-1222683904

I should also note that `w3.org` URIs are preferred over `w3id.org` URIs in many WG-related scenarios, and while administration of `w3.org` redirects requires going through additional bureaucracy and lacks the simplicity of `w3id.org` updates via git, a functionally equivalent permalink-redirect arrangement should be possible.

pchampin:
@ioggstream 
> I just meant that I expected YAML-LD to be interchangeable with JSON-LD without the need of bearing specific knowledge around.

and that's why, in my view, this "convenience context" is the right way to go. It simply relies on a standard JSON-LD mechanism.

Start with the following YAML-LD:
```yaml
"@context": http://www.w3.org/ns/yaml-ld/v1
$id: http://champin.net/#pa
$type: http://schema.org/Person
```
Convert it to JSON(-LD)
```json
{
  "@context": "http://www.w3.org/ns/yaml-ld/v1",
  "$id": "http://champin.net/#pa",
  "$type": "http://schema.org/Person"
}
```
which then expands to 
```json
{
  "@id": "http://champin.net/#pa",
  "@type": "http://schema.org/Person"
}
```
and converts to
```
<http://champin.net/#pa> a <http://schema.org/Person>.
```

As @gkellogg pointed out above, it could even be used in other JSON-LD documents where using `$` instead of `@` would be considered desirable. Conversely, its use would not be required by YAML-LD, and YAML-LD processors would still recognize good old `@` keywords.

spec/index.html
gkellogg: Fixed in 76ceeb26a770d522e64b9173158242a11d4a1ea0.

spec/index.html
gkellogg: Fixed in 76ceeb26a770d522e64b9173158242a11d4a1ea0.

spec/index.html
gkellogg: Issues surrounding accessing remote context continues to be a hot topic in JSON-LD (driven by schema.org) and should have a fair amount of discussion at TPAC, probably in the later portion of our F2F time.

anatoly-scherbakov (approved):

spec/index.html
BigBlueHat: On a previous call we'd discussed a "convenience context" but my understanding was that we would simply be providing a context which removed the `@` symbol...not "blessing" a new symbol with magic meaning... Also, given that JSON Schema uses the `$` prefix heavily, attempting to us that runs us into a namespace collision with them (of sorts) and won't make us any friends.

I could support providing a simple context file that aliases things without the `@` prefix (i.e. `"base": "@base`), but developers will have to be aware (as ever) of what that means for the rest of their document/contents.

gkellogg:
[On today's meeting](https://json-ld.org/minutes/2022-08-31/#resolution-1) we [resolved](https://json-ld.org/minutes/2022-08-31/#resolution-1) to postpone this and include something similar in a Best Practices document.

> turn PR yaml-ld#77 to draft, and consider move this to Best Practices

There was further discussion about using a simpler bare-word Connivence Context, such as https://github.com/json-ld/convenience-context which we might host at https://json-ld.org/contexts/.

PR #99, opened 2023-05-04
#97: Document Extended → Basic conversion

anatoly-scherbakov:


spec/index.html
```diff
@@ -2383,12 +2381,127 @@ <h4>YamlLdErrorCode</h4>
...
+        </p>
+
+        <p>
+          Instead, we implement two API functions:
+        </p>
+
+        <dl data-sort>
+          <dt><code>extended_to_basic(extended_document: YAML-LD) → YAML-LD</code></dt>
+          <dd>
+            Converts an Extended YAML-LD document to a JSON-LD document that a standard
```
TallTed, line 2408: This changes the `<dd>` content to be in sync with the `<dt>` and `<h2>` containing it. 
```suggestion
          <dt><code>extended_to_basic(extended_document: YAML-LD) → YAML-LD</code></dt>
          <dd>
            Converts an Extended YAML-LD document to a YAML-LD document that a standard
```

spec/index.html
```diff
@@ -2383,12 +2381,127 @@ <h4>YamlLdErrorCode</h4>
...
+            processor can work with
+          </dd>
+
+          <dt><code>basic_to_extended(basic_document: JSON-LD | YAML-LD) → YAML-LD</code></dt>
+          <dd>
+            Converts a basic JSON-LD or YAML-LD document to the Extended form, making
```
TallTed, line 2414: This needs some discussion of how JSON-LD gets involved in "converting Extended YAML-LD to Basic YAML-LD and back".

spec/index.html
```diff
@@ -2383,12 +2381,127 @@ <h4>YamlLdErrorCode</h4>
           </section>
         </section>
 
+        <section>
+          <h2>Implementations</h2>
```
gkellogg, line 2385: Implementations are typically listed to the Implementation Report, which should be linked to the spec.

spec/index.html
```diff
@@ -2383,12 +2381,127 @@ <h4>YamlLdErrorCode</h4>
...
+          This approach is simpler than the Extended Internal Representation because it
+          does not require any changes to the internal structures of existing JSON-LD
+          libraries.
+        </p>
+
+        <p>
+          Instead, we implement two API functions:
+        </p>
+
+        <dl data-sort>
```
gkellogg, line 2405: To prevent very complex processing model, it's important that these functions work on the expanded representation, not compacted. Otherwise, keeping track of term definitions and scoped contexts is quite difficult.

spec/index.html
gkellogg: It shouldn't need to go through JSON.

spec/index.html
gkellogg: Also, we'll need to clarify that this is not really a round-trip, as it can't know the specifics of the original syntax usage, rather it "optimizes"  a basic YAML-LD document using extended features.

spec/index.html
anatoly-scherbakov: Yes, but these implementations are specifically for YAML-LD Extended profile and they do not relate to the body of the spec, do they?

We could also have a separate section for YAML-LD Extended → YAML-LD simple conversion implementations.

spec/index.html
```diff
@@ -2383,12 +2381,155 @@ <h4>YamlLdErrorCode</h4>
...
+          <dt><code>extended_to_basic(extended_document: YAML-LD) → YAML-LD</code></dt>
+          <dd>
+            <ul>
+              <li>Performs JSON-LD expansion to the given Extended YAML-LD document</li>
```
TallTed, line 2411: I still don't understand what JSON-LD is doing in a YAML-LD-to-YAML-LD transformation.

spec/index.html
```diff
@@ -2383,12 +2381,155 @@ <h4>YamlLdErrorCode</h4>
...
+              <li>Converts it to Basic YAML-LD form</li>
+            </ul>
+          </dd>
+
+          <dt><code>basic_to_extended(basic_document: YAML-LD) → YAML-LD</code></dt>
+          <dd>
+            <ul>
+              <li>Performs JSON-LD expansion to the given Basic YAML-LD document</li>
```
TallTed, line 2419: I still don't understand what JSON-LD is doing in a YAML-LD-to-YAML-LD transformation.

spec/index.html
gkellogg: It's necessary to be sure that the input document is in expanded form so that we don't need a mechanism parallel to JSON-LD expansion to properly handle all the term types, containers, and scoped contexts. Doing this in expanded form prevents what would otherwise be a very complicated process, in the general case.

TallTed: Something like this explanation should be put in the document, including how and why "JSON-LD expansion" is applied to a YAML-LD document, which isn't JSON-LD, which makes it very confusing that "JSON-LD expansion" is applied instead of "YAML-LD expansion".

anatoly-scherbakov: Hmm, this is puzzling. I see the following two alternatives how we can word this.

1. We just specify that source YAML-LD document has to be expanded beforehand.
   * but then, we have to specify what YAML-LD expansion is,
   * and the interface of the functions becomes not entirely user friendly: user has to _remember_ that the document has to be expanded, and the functions probably should _verify_ that.

2. Do the expansion ourselves:
   * Convert YAML → JSON,
   * Perform JSON-LD Expansion,
   * Convert resulting JSON back to YAML,
   * Perform the necessary basic → extended or extended → basic conversion.

I would lean to the option (2).

Of course, internally, when we're using `rdflib` for instance, — we won't actually convert YAML to JSON and back; instead we will read YAML into an internal Python data structure of nested dicts, lists, strings, numbers and `None`s, which `rdflib` will subsequently process. But I doubt we can say it that way in the spec.

spec/index.html
anatoly-scherbakov: Commented above on that :white_check_mark:

spec/index.html
TallTed: I think it's fine to say something like "You won't typically need to perform these steps manually, because libraries such as `rdflib` will take care of them under the covers, but it can help with troubleshooting and optimization to know what's going on. So, you start with YAML, convert it to JSON, perform JSON-LD Expansion, convert that to YAML-LD, and do any necessary basic → extended or extended → basic conversion on the YAML-LD. Alternatively, your library might do YAML-LD expansion directly on the initial YAML document, and then do any necessary basic → extended or extended → basic conversion on the YAML-LD."

anatoly-scherbakov: Thanks @TallTed! I added this paragraph.

Also, I think we actually can't convert Extended → Basic exactly through JSON-LD because Extended YAML-LD specifically can't be convert to JSON-LD. Right?

gkellogg: You'll have to convert YAML to the YAML JSON profile, which is compatible.

anatoly-scherbakov: What about like this? Should extended → basic require conversion?

How do scoped contexts and stuff influence the conversion of type tags to `@type` declarations?

spec/index.html
```diff
@@ -2425,6 +2427,19 @@ <h2>Convert Extended YAML-LD to Basic YAML-LD and back</h2>
           </dd>
         </dl>
 
+        <div class="note">
+          ou won't typically need to perform these steps manually
```
TallTed, line 2430:
```suggestion
          You won't typically need to perform these steps manually
```

spec/index.html
```diff
@@ -2425,6 +2427,19 @@ <h2>Convert Extended YAML-LD to Basic YAML-LD and back</h2>
...
+          because libraries such as <code>rdflib</code> will take care of them
+          under the covers, but it can help with troubleshooting and
+          optimization to know what's going on. So, you start with YAML,
+          convert it to JSON, perform JSON-LD Expansion, convert that to
+          YAML-LD, and do any necessary basic → extended or extended → basic
+          conversion on the YAML-LD. Alternatively, your library might do
+          YAML-LD expansion directly on the initial YAML document, and then
+          do any necessary basic → extended or extended → basic conversion on
```
TallTed, line 2438:
```suggestion
          YAML-LD, and do any necessary <em>basic → extended</em> or <em>extended → basic</em>
          conversion on the YAML-LD. Alternatively, your library might do
          YAML-LD expansion directly on the initial YAML document, and then
          do any necessary <em>basic → extended</em> or <em>extended → basic</em> conversion on
```

TallTed (changes requested):
Fix a typo, and add some `<em>` for clarity

Issue #94, opened 2023-03-20
Unify YAML-LD & JSON-LD examples


anatoly-scherbakov:
## Why

Examples in the current spec document are very differently themed.

## What

Convert all code examples into one unified theme throughout the whole specification.

## Which theme might that be?

Proposing to use the theme of the specification itself. For instance, we can describe its title, the fact that it derives from JSON-LD, the fact that it uses YAML as its syntax, etc.

gkellogg:
Looks like a good approach to me.

PR #114, opened 2023-08-12
Remove a few tests (PR #7 for issue #94)

anatoly-scherbakov:
# Alias nodes are allowable

Even for the Basic profile.

# Questions

* I do not think we are running tests in CI. Are we?
* Should we? Or each implementation should care about them itself?

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/114.html" title="Last updated on Aug 17, 2023, 8:19 AM UTC (02fcabe)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/114/a82183a...02fcabe.html" title="Last updated on Aug 17, 2023, 8:19 AM UTC (02fcabe)">Diff</a>


gkellogg:
Running tests in CI would need to choose some specific implementation, and is prone to false positives/negatives if that implementation is incorrect. Usually, problems are either found when an implementation encounters them, or when there's a disagreement when the test is submitted or the report run.

gkellogg (approved):

PR #120, opened 2023-09-03
A few `basic` tests failing

anatoly-scherbakov:
# Comments test

Expand operation converts a mapping with a # in key to an empty object. Try this out in the playground:

```json
{
  "@context": {
    "@vocab": "http://example.com/"
  },
  "@id": "http://example.org/test#example",
  "@type": "t1",
  "notacomment": [
    {"this is#not": "a comment"}
  ]
}
```

### DIFF
--- a/tests/basic/cir-scalar-other-1-positive-out.yamlld
+++ b/tests/basic/cir-scalar-other-1-positive-out.yamlld
@@ -7,6 +7,8 @@
   - "@value": 123.45
   http://example.com/ystring:
   - "@value": string
+  http://example.com/ynull:
+  - "@value": null
   http://example.com/ybool:
   - "@value": true
   http://example.com/integer:
--- a/tests/basic/cir-scalar-other-1-positive-out.yamlld
+++ b/tests/basic/cir-scalar-other-1-positive-out.yamlld
@@ -7,8 +7,6 @@
   - "@value": 123.45
   http://example.com/ystring:
   - "@value": string
-  http://example.com/ynull:
-  - "@value": null
   http://example.com/ybool:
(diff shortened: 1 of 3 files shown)


TallTed (approved):
lgtm

tests/basic/cr-comments-1-positive-out.yamlld
```diff
@@ -3,8 +3,7 @@
   "@type":
   - http://example.com/t1
   http://example.com/notacomment:
-  - http://example.com/this is#not:
-    - "@value": a comment
+  - {}  # That is what we get from `this is#not: a comment`. This behavior is confirmed by JSON-LD playground.
```
gkellogg, line 6: I believe the original is correct, even if the playground gets a different value. "this is#not" should be treated as an IRI, relative to "http://example.com/". The fact that it is an illegal IRI is not something that the JSON-LD Expansion algorithms check for. But, arguably, we should have positive test results that contain illegal IRIs. It might be better to use something like `"this-is#not": a comment`.

tests/basic/cir-scalar-other-1-positive-out.yamlld
```diff
@@ -7,6 +7,8 @@
   - "@value": 123.45
   http://example.com/ystring:
   - "@value": string
+  http://example.com/ynull:
+  - "@value": null
```
gkellogg, line 11: JSON-LD processors drop `null` values, so this would not be a valid expansion result.

gkellogg (changes requested):
In short, I think the existing values are correct, although arguably the comment one could be changed to make the results valid.

PR #142, opened 2024-04-16
#140 Replace `extendedYAML` with `processingMode`

### LINKED ISSUE #140: `extendedYAML` flag clashes with `processingMode` option
# Why

JSON-LD 1.1 API spec [states](https://w3c.github.io/json-ld-api/#dom-jsonldoptions-processingmode) the following about `processingMode` parameter:

> If set to json-ld-1.0 or json-ld-1.1, the implementation must produce exactly the same results as the algorithms defined in this specification. If set to another value, the JSON-LD processor is allowed to extend or modify the algorithms defined in this specification to enable application-specific optimizations. The definition of such optimizations is beyond the scope of this specification and thus not defined. Consequently, different implementations may implement different optimizations. Developers must not define modes beginning with json-ld as they are reserved for future versions of this specification.

YAML-LD spec in its non-normative part [defines](https://json-ld.github.io/yaml-ld/spec/#dom-jsonldoptions-extendedyaml) the `extendedYAML` flag:

> ...
> If set, allows the use of [node tags](https://yaml.org/spec/1.2.2/#node-tags) when serializing [RDF literal](https://www.w3.org/TR/rdf11-concepts/#dfn-literal) values having datatypes other than xsd:string or [language-tagged strings](https://www.w3.org/TR/rdf11-concepts/#dfn-language-tagged-string) as [scalar](https://yaml.org/spec/1.2.2/#scalar) values.
> Otherwise, it serializes [RDF literal](https://www.w3.org/TR/rdf11-concepts/#dfn-literal) values to the closest [scalar](https://yaml.org/spec/1.2.2/#scalar) representation from the [YAML Core Schema](https://yaml.org/spec/1.2.2/#core-schema).

I feel we might have a duplication.

# Proposal

What if we write

```json
{
  "processingMode": "yaml-ld-extended"
}
```

instead?

# Consequences

* :heavy_plus_sign:  We reduce the complexity and reuse the existing functionality with conformance to existing specifications;
* :heavy_plus_sign:  We do not have to **delete** the `processingMode` option as per https://github.com/json-ld/yaml-ld/issues/139 - instead,
  * We will keep it for the future,
  * But specify that `json-ld-1.0` value MUST cause the processor to throw an error :boom: .

anatoly-scherbakov:
# Why

`extendedYAML` might be replaced by using `processingMode`.

# What 

Closes #140.

***
<a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/pull/142.html" title="Last updated on Apr 16, 2024, 7:56 PM UTC (0448b40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/json-ld/yaml-ld/142/f5b169d...0448b40.html" title="Last updated on Apr 16, 2024, 7:56 PM UTC (0448b40)">Diff</a>

### DIFF
--- a/spec/index.html
+++ b/spec/index.html
@@ -616,7 +616,7 @@ schema:hasPart:
 
     <ul>
       <li>
-        <a href="#dom-jsonldoptions-processingmode">processingMode</a> option is provided and set to `json-ld-1.0`.
+        <a href="#dfn-processingmode">processingMode</a> option is provided and set to `json-ld-1.0`.
       </li>
     </ul>
 
@@ -1627,7 +1627,7 @@ schema:hasPart:
         </p>
 
         <p id="cir-extended" data-tests="">
-          If the {{JsonLdOptions/extendedYAML}} API flag is `true`, the processing result
+          If the {{JsonLdOptions/processingMode}} API parameter is `yaml-ld-extended`, the processing result
           will be in the <a>extended internal representation</a>.
         </p>
 
@@ -1875,7 +1875,7 @@ schema:hasPart:
             <h3>Converting a <a>YAML scalar</a></h3>
 
             <ol>
-              <li>If the {{JsonLdOptions/extendedYAML}} flag is `true`,
+              <li>If the {{JsonLdOptions/processingMode}} option is `yaml-ld-extended`,
                 and <a>node</a> |n|
                 has a <a>node tag</a> |t|,
                 |n| is mapped as follows:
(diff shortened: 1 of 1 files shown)


spec/index.html
```diff
@@ -2359,9 +2359,7 @@ <h3>JsonLdProcessor</h3>
                 <a data-cite="JSON-LD11-API#rdf-to-object-conversion">RDF to Object Conversion</a> algorithm
                 before step 2.6 as follows:
                 <blockquote>
-                  Otherwise, if both the {{JsonLdOptions/useNativeTypes}}
-                  and {{JsonLdOptions/extendedYAML}} flags are set
-                  and the
+                  Otherwise, if both the {{JsonLdOptions/useNativeTypes}} flag is set and {{JsonLdOptions/processingMode}} parameter is `yaml-ld-extended` and the
                   <a>datatype IRI</a>
```
TallTed, line 2363:
```suggestion
                  Otherwise, if the {{JsonLdOptions/useNativeTypes}} flag is set,
                  the {{JsonLdOptions/processingMode}} parameter is `yaml-ld-extended`,
                  and the <a>datatype IRI</a>
```

TallTed (changes requested):

gkellogg (approved):

PR #154, opened 2024-06-30
➕ `#frame-t0001`

anatoly-scherbakov:


anatoly-scherbakov:
closes: https://github.com/json-ld/yaml-ld/issues/148

tests/cases/frame/0001-out.yamlld
```diff
@@ -0,0 +1,16 @@
+"@context":
+  dcterms: http://purl.org/dc/terms/
+  ex: http://example.org/vocab#
+"@graph":
```
gkellogg, line 4: My implementation does not place the result inside of `@graph`. I get the following:

```yaml
---
"@context":
  dcterms: http://purl.org/dc/terms/
  ex: http://example.org/vocab#
"@id": http://example.org/test/#library
"@type": ex:Library
ex:contains:
  "@id": http://example.org/test#book
  "@type": ex:Book
  dcterms:contributor: Writer
  dcterms:title: My Book
  ex:contains:
    "@id": http://example.org/test#chapter
    "@type": ex:Chapter
    dcterms:description: Fun
    dcterms:title: Chapter One
```

This is because `@graph` is only used in the output if the input frame uses `@graph`, or the `frameDefault` option is set to `true`. See step 13 of the [frame() API method](https://www.w3.org/TR/json-ld11-framing/#dom-jsonldprocessor-frame).

gkellogg (changes requested):

tests/cases/frame/0001-out.yamlld
anatoly-scherbakov: Thanks! Fixed.

tests/cases/frame/0001-out.yamlld
```diff
@@ -0,0 +1,7 @@
+"@context":
+  dcterms: http://purl.org/dc/terms/
+  ex: http://example.org/vocab#
+"@id": http://example.org/test/#library
+"@type": ex:Library
+ex:contains:
+  "@id": http://example.org/test#book
```
gkellogg, line 7: Actually, it should frame the whole document recursively:

```suggestion
  "@id": http://example.org/test#book
  "@type": ex:Book
  dcterms:contributor: Writer
  dcterms:title: My Book
  ex:contains:
    "@id": http://example.org/test#chapter
    "@type": ex:Chapter
    dcterms:description: Fun
    dcterms:title: Chapter One
```

gkellogg (changes requested):

gkellogg (approved):

PR #160, opened 2025-04-01
Describe vocabulary prefixes used in spec

anatoly-scherbakov:


spec/data/spec.yamlld
```diff
@@ -14,6 +14,9 @@ license:
   $id: https://purl.org/NET/rdflicense/W3C1.0
   dcterms:isReferencedBy: https://purl.org/NET/rdflicense/W3C1.0.ttl
 
+dcterms:isReferencedBy:
```
pchampin, line 17: is this the appropriate predicate?
I don't see any reference to the YAML-LD spec in namespace-prefixes.yamld...

TallTed (commented):
Nothing bad leaps out at me. But I'm not quite fluent in YAML nor YAML-LD.

spec/data/spec.yamlld
anatoly-scherbakov: Makes sense. What about `dcterms:requires`?

pchampin: I don't think that the spec *requires* the prefix definitions... They are used in examples only.

If anything, dcterms:references would work, IMO.

anatoly-scherbakov: Done :white_check_mark:

pchampin (approved):

PR #161, opened 2025-04-19
More inline LD about spec authors

anatoly-scherbakov:
image

### DIFF
--- a/spec/data/spec.yamlld
+++ b/spec/data/spec.yamlld
@@ -19,6 +19,7 @@ schema:author:
     $type: foaf:Person
     schema:givenName: Ted
     schema:familyName: Thibodeau
+    schema:honorificSuffix: Jr
 
   - $id: https://github.com/BigBlueHat
     $type: foaf:Person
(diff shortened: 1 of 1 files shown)


spec/data/spec.yamlld
```diff
@@ -4,29 +4,46 @@ schema:author:
...
+
+  - $id: https://github.com/ioggstream
+    $type: foaf:Person
+    schema:givenName: Roberto
+    schema:familyName: Polli
+
+  - $id: https://github.com/TallTed
+    $type: foaf:Person
+    schema:givenName: Ted
+    schema:familyName: Thibodeau
```
TallTed, line 21: I don't know whether this is better
```suggestion
    schema:familyName: Thibodeau Jr
```

TallTed: or this is better
```suggestion
    schema:familyName: Thibodeau
    schema:honorificSuffix: Jr
```

TallTed: [schema.org](https://schema.org/Person) doesn't mention Jr, Sr, III, etc., so I'm not sure where to put my `Jr`.

Issue #83, opened 2022-09-01
use regular expressions to resolve literals; URIs, CURIEs


VladimirAlexiev:
As a data architect
I want YAML plain values to be recognized by regular expression
So that I don't have to explicitly tag them

- The YAML spec https://yaml.org/spec/1.2.2/#332-resolved-tags says
"Application specific tag resolution rules should be restricted to resolving the “?” non-specific tag, most commonly to resolving plain scalars. These may be matched against a set of regular expressions to provide automatic resolution of integers, floats, timestamps and similar types."
- The default "JSON schema" https://yaml.org/spec/1.2.2/#102-json-schema includes regexes for matching integers and floats
- The extended "Core schema" https://yaml.org/spec/1.2.2/#103-core-schema includes regexes for matching integers, floats, NaNs, null, booleans
- The https://yaml.org/type collection lists further interesting cases:
  - https://yaml.org/type/timestamp.html for timestamps (such examples used to be in the spec, but are being deprecated)
  - https://yaml.org/type/value.html for extensibility of strings with extra attributes
- We could define regexes for matching URNs and IRIs instead of tagging them explicitly #79

Examples (Tagging @OR13 and @mgh128 who work with EPCIS data):
```yaml
"@context": 
  epcis: https://ns.gs1.org/epcis/
issued: 2022-09-01
stringThatOnlyLooksLikeADate: !string 2022-09-01
homepage: https://example.org/foo
stringThatOnlyLooksLikeAUrl: !string https://example.org/foo
urlThatMayBeMisspelled: !anyURI hxxp:\\i-cannot spell,con/my home page
epcis:readPoint: urn:epc:id:sgln:952005385.011.ts4711
epcis:epcList: https://id.gs1.org/01/70614141123451/21/2018
```

Note: the benefit of datatype `xsd:anyURI` (tag `!anyURI`) is that:
- the URL may be misspelled, because it is stored as a literal and semantic repos (at least rdf4j) don't check its syntax. How's that a benefit? Try to import a million CrunchBase "homepage" props and you'll see
- OWL ontologies frown upon ObjectProperties that don't lead to any triples (eg rdf:type owl:NamedIndividual and some others)

We could also use explicit delimiters eg `<...>` around URNs (URIs, IRIs), which will also enable the use of [CURIEs](https://www.w3.org/TR/curie/).
Eg below each of the props `epcis:readPoint, epcis:epcList` has 2 identical values (first a full URN, then a CURIE), 
[...]

PR #174, opened 2026-02-22
#83 YAML-LD is based on YAML Core schema (with exceptions)

anatoly-scherbakov:
# Why

What values are recognized and what are not?

# What

Let's specify that we rely upon YAML 1.2.2 Core Schema (with a few exceptions).

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/pull/174.html" title="Last updated on Mar 15, 2026, 3:27 PM UTC (9ee1210)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/174/8af4a28...9ee1210.html" title="Last updated on Mar 15, 2026, 3:27 PM UTC (9ee1210)">Diff</a>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ spec:
 .PHONY: serve
 .ONESHELL: serve
 serve:
-	python -m http.server --directory web 1234
+	python -m http.server --directory . 1234
 
 .PHONY: html-manifest
 .ONESHELL: html-manifest


tests/cases/value-types/id-type-bool-yes.yamlld
ioggstream: Currently, the behavior of this document depends on the YAML parser version, so we should add it in the testcase.

```yaml
%YAML 1.2
---
...
```

I think that for YAML 1.2, `Yes` and `no' are strings, so valid `@id`s.

Note that pyyaml uses YAML 1.1, and interprets `no` as `false`.

ioggstream (approved):
Ooop, commented the wrong PR.

PR #202, opened 2026-05-17
[#201] Fix broken documentLoader links

### LINKED ISSUE #201: Broken links in YAML-LD 1.0
While crawling [YAML-LD 1.0](https://w3c.github.io/yaml-ld/), the following links to other specifications were detected as pointing to non-existing anchors:
* [ ] https://www.w3.org/TR/json-ld11/#dom-jsonldoptions-documentloader

<sub>This issue was detected and reported semi-automatically by [Strudy](https://github.com/w3c/strudy/) based on data collected in [webref](https://github.com/w3c/webref/).</sub>

anatoly-scherbakov:
## Summary

- Fix `documentLoader` citations to point at JSON-LD 1.1 API instead of JSON-LD 1.1 Syntax.
- Validate both ReSpec documents in CI with link validation enabled.
- Add project guidance for the Makefile ReSpec validation targets.

## Validation

- `npx -y respec --verbose --localhost --src index.html --out web/index.html`
- `npx -y respec --verbose --localhost --src extended-profile/index.html --out web/extended-profile/index.html`
- Confirmed generated output links to `https://www.w3.org/TR/json-ld11-api/#dom-jsonldoptions-documentloader` and does not contain the broken `json-ld11/#dom-jsonldoptions-documentloader` link.
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml")'`
- `git diff --check`

Closes #201

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/pull/202.html" title="Last updated on May 20, 2026, 8:41 PM UTC (0a7e158)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/202/2e3d53c...0a7e158.html" title="Last updated on May 20, 2026, 8:41 PM UTC (0a7e158)">Diff</a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,17 +9,8 @@ on:
 
 jobs:
   tests:
-    name: Build, Validate and Deploy (${{ matrix.source }})
+    name: Build, Validate and Deploy
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        include:
-          - source: index.html
-            specStatus: WD
-            shortName: yaml-ld-10
-          - source: extended-profile/index.html
-            specStatus: DNOTE
-            shortName: yaml-ld-extended-profile
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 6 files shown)


pchampin (changes requested):
* the changes in the specs look good
* I object to the changes in .github/workflows/ci.yml
   I would rather have two separate files, once for yaml-ld, one for the extended profile note.
   I'm usually not a big fan of copy-paste, but Echidna is a powerful tool with impact on /TR, so I'm a bit uncomfortable making a "blanket" configuration that can easily be misused by just tweaking the 'matrix' parameter.

   When using Echidna in a repo containing multiple specs, good practice consists in filtering workflows by subdirectory (see for example: https://github.com/w3c/epub-specs/blob/main/.github/workflows/epub-anno.yml).
  I appreciate that its is not directly applicable here, because the YAML-LD spec is at the root. I would suggest that we either
  + bite the bullet and make this repo properly "multi-spec" with one folder per spec, or
  +  stick to one-repo-per-spec, and move the extended profile to another repo.

* I'm not sure the AGENT.md CLAUDE.md files should be committed -- all the more that they reference a Makefile that is not committed.

anatoly-scherbakov:
Thanks, I addressed the CI concern by removing the matrix-based spec-prod setup. The root YAML-LD spec is back in its own workflow, and the extended profile now has a separate validation-only workflow without Echidna publishing.

On `AGENTS.md` / `CLAUDE.md`: I would prefer to keep them in this PR. The `Makefile` they reference is already committed in the repository on `main`, so the guidance points agents to existing project commands rather than introducing an uncommitted dependency. Agent-driven development is increasingly becoming part of routine maintenance, and repository-local guidance helps agents act adequately here: in particular, it tells them to use the project ReSpec validation targets when touching the specs.

pchampin (approved):
about 'Makefile': my bad, it is indeed there.
LGTM now

PR #206, opened 2026-06-07
[#94]: Unify spec examples under Proxima Centauri domain

### LINKED ISSUE #94: Unify YAML-LD & JSON-LD examples

anatoly-scherbakov:
## Motivation

Examples in the spec follow different topics: John Lennon, Moon, whatever. Most of the examples are self referential, they talk about the spec itself, which is not very ergonomic since, if the spec changes, the examples should also change.

Let's use a unified theme for all examples: the star system containing the nearest exoplanet to Earth we know of.

## Summary

- Rewrites all spec examples (`intro`, `basic`, `stream`, `json-literal`, `html-embed`, `context-anchors`) to use a single coherent semantic topic: the Proxima Centauri star system
- Replaces self-referential and ad-hoc example subjects (John Lennon, the spec itself, Moon) with real DBpedia-backed entities using `dbr:`, `dbo:`, `dbp:` prefixes and canonical `http://dbpedia.org/` IRIs
- Deletes `anchors.yamlld` (semantically wrong `schema:hasPart` relationship); promotes `context-anchors.yamlld` to the primary anchors example, rewritten to show `&iri` anchor on Proxima Centauri b discovery properties
- `basic.jsonld` is now generated from `basic.yamlld` via `pyld get`; `context-anchors.jsonld` likewise generated and shown in the spec as the result block
- Pins `iolanta>=2.1.36` (bug fix release for non-editable installs)

## Test plan

- [ ] `make generate` produces identical output (no git diff after running)
- [ ] `make spec` renders without errors
- [ ] All example URIs dereference to labeled DBpedia resources (validated with `iolanta kglint/json`)

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/pull/206.html" title="Last updated on Jun 13, 2026, 8:04 PM UTC (3cc330d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/206/d51465d...3cc330d.html" title="Last updated on Jun 13, 2026, 8:04 PM UTC (3cc330d)">Diff</a>


TallTed (commented):
lgtm

Issue #220, opened 2026-07-12
Reference YAML processing diagram in spec structure


anatoly-scherbakov:
> We could go further by mentioning the different steps from Figure 1 in the remaining subsections. E.g. 4.1 Encoding
> is about the 'Present' and 'Parse' steps. §4.2 Comments and §4.3 Anchors and Aliases directly follow from the overall processing: comments are not part of the intermediary structures (besides 'Presentation'), and anchors/aliases are "expanded" in the 'Representation' structure, so clearly they can not be leveraged by JSON-LD processing. etc...

© @pchampin at https://github.com/w3c/yaml-ld/pull/219

PR #241, opened 2026-07-28
Restore Basic-profile normative requirements stranded after #220

### LINKED ISSUE #230: Restore Basic-profile normative requirements stranded after #220
# Why

After the YAML processing-stages restructure (#220), several **Basic-profile** normative requirements (and the fragment IDs the test suite links to) live only in the non-normative Extended Profile note. The main ED (`index.html`) no longer defines them, but `tests/manifest.html` still points at `../index.html#…` for those anchors.

Basic conformance must not depend on a DNOTE. This is a Rec-track correctness blocker. Extended also still says Basic streams MUST be a single document, which contradicts the main multi-doc + `extractAllScripts` model.

# What

Restore into `index.html` under Core Requirements / Native Data Structure (stable IDs), drawing from Extended / the March WD:

- `#cr-well-formed` — streams MUST be well-formed; aliases MUST resolve to a prior anchored node; else `loading-document-failed`
- `#cir-document-content` — top-level document node MUST be a mapping or sequence
- Non-Core / unknown `node tags` — MUST reject for Basic (fail load); close the gap that leaves `!!binary` / language tags implementation-defined
- `#stream-processing` — default for `extractAllScripts` on `application/ld+yaml` (JSON-LD API default `false`); empty-stream behavior
- Minimal **dump** requirements — UTF-8; string keys; JSON-compatible scalars; semantic round-trip only (not comments/anchors/key order)
- Optionally: large integers beyond JSON safe integer range (fail vs preserve)
- **Mapping keys (Option 1, locked):** keys MUST be Core Schema `tag:yaml.org,2002:str` in the **representation graph**; post-Construct stringification (e.g. JS turning `1:` into `"1"`) does not make a non-string key conforming; `mapping-key-error` MUST abort (`data-link-for="YamlLdErrorCode"`)

Retarget broken test anchors in `tests/manifest.html` (among others: `#cr-well-formed`, `#cir-stream`, `#cir-document`, `#cir-document-content`, `#cir-mapping`, `#cir-mapping-key`, `#cir-scalar-*`, `#cir-alias-json`, `#cir-alias-cycles`, `#jp-utf8`, `#ep-utf8`).

In Extended: keep Convert algorithms as informative only; remove/rewrite single-document MUST for Basic (full Extended cleanup may finish in the [Extended sync issue (#237)](https://github.com/w3c/yaml-ld/issues/237)).

Out of scope here: [boolean mapping-key test (#231)](https://github.com/w3c/yaml-ld/issues/231); [Extended typos/biblio (#237)](https://github.com/w3c/yaml-ld/issues/237).

Acceptance: IDs restored; Option 1 mapping-key wording in ED; test hrefs resolve in `index.html`; `make spec` passes.

anatoly-scherbakov:
Closes #230.

Restores Basic-profile requirements, removes the Extended single-document contradiction, and generates test-manifest references from both specifications.

Validated with:
- `npm run generate`
- `make spec`
- `make extended-profile`
- Playwright cross-specification link and orphan-test audit

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/pull/241.html" title="Last updated on Jul 28, 2026, 7:47 PM UTC (ca054b6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/241/7b0eb1e...ca054b6.html" title="Last updated on Jul 28, 2026, 7:47 PM UTC (ca054b6)">Diff</a>


BigBlueHat:
Thanks, @anatoly-scherbakov. I'm still reviewing. However, I did find [Example 6](https://pr-preview.s3.amazonaws.com/w3c/yaml-ld/241/7b0eb1e...ca054b6.html#example-non-string-mapping-key-raises-mapping-key-error) pretty confusing...

There's too much going on in there (dollar sign aliases, the `*iri` tag, etc.).

Could we simplify that example to focus on what within it will throw the mapping key error?

Thanks!
🎩

anatoly-scherbakov:
@BigBlueHat thanks for the review! Since this PR does not add that example I created a separate issue to tackle it.

index.html
```diff
@@ -1157,12 +1182,24 @@ <h2>Scalar Value Types</h2>
...
+        <p id="node-tags">
+          A <a>YAML scalar</a> whose resolved <a>node tag</a> is outside the
+          <a data-cite="YAML#103-core-schema">YAML Core Schema</a>
+          MUST cause a
+          <a data-cite="JSON-LD11-API#dom-jsonlderrorcode-loading-document-failed">loading-document-failed</a>
+          error to be raised when constructing the JSON-LD [=internal representation=].
+        </p>
+
+        <p class="note">
+          The informative [[[yaml-ld-extended-profile]]] describes alternate handling of such tags.
```
BigBlueHat, line 1200: @anatoly-scherbakov I think we should not use `[[[]]]` here and just raw links. This won't likely be a TR-space document. Just a `Note`, I'd expect. Consequently, a normal old `<a href="">` would avoid Respec expecting (and minting) a TR-space URL as it does now.

So...

```suggestion
          The informative <a href="../extended-profile/">yaml-ld-extended-profile</a> describes alternate handling of such tags.
```

That should do the trick.

BigBlueHat: /cc @pchampin wrt to TR-space and Note statements.

BigBlueHat (changes requested):

index.html
BigBlueHat: Ignore this. Overtaken by the group resolution to publish this as a Note.

BigBlueHat (approved):

extended-profile/index.html
```diff
@@ -1182,13 +1175,6 @@ <h3>YAML-LD Extended Profile</h3>
...
 
-            <p>
-              As with the <a>YAML-LD Basic profile</a>,
-              a <a>YAML stream</a> MUST include only a single <a>YAML document</a>,
-              as the <a>JSON-LD extended internal representation</a> only supports
-              a single document model.
-            </p>
-
             <p class="issue" data-number="79">
               Consier something like `!id` as a local tag to denote IRIs.
```
TallTed, line 1179:
```suggestion
              Consider something like `!id` as a local tag to denote IRIs.
```

tests/manifest.html
```diff
@@ -273,7 +273,7 @@ <h1>Test Definitions</h1>
                   <dt>Name</dt><dd>language-tagged string with i18n (ignored without extendedYAML)</dd></dt>
```
TallTed, line 273: Is `i18n` translated anywhere? It might be better to change all instances of this insider "word" to `internationalization`.

